PR #5, opened 2021-02-13
Add units to Logger

Piphi5:


src/main/native/cpp/view/Logger.cpp
```diff
@@ -74,6 +66,56 @@ void Logger::Display() {
...
+  // Add Units Per Rotations entry
+
+  // Rotational units have fixed Units per rotations
+  m_isRotationalUnits =
+      (m_settings.units == "Rotations" || m_settings.units == "Degrees" ||
+       m_settings.units == "Radians");
+  if (m_settings.units == "Degrees") {
+    m_settings.unitsPerRotation = 360.0;
+  } else if (m_settings.units == "Radians") {
+    m_settings.unitsPerRotation = 2 * M_PI;
```
prateekma, line 104: Use wpi::math::pi from <wpi/math>

src/main/native/cpp/view/Logger.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "sysid/view/Logger.h"
 
+#include <cmath>
```
prateekma, line 7: Remove this include after fixing the comment above.

prateekma (changes requested):

src/main/native/cpp/view/Logger.cpp
```diff
@@ -74,6 +66,56 @@ void Logger::Display() {
...
+  // Add Dropdown for Units
+  ImGui::SetNextItemWidth(ImGui::GetFontSize() * 10);
+  if (ImGui::Combo("Unit Type", &m_selectedUnit, kUnits,
+                   IM_ARRAYSIZE(kUnits))) {
+    m_settings.units = kUnits[m_selectedUnit];
+  }
+
+  sysid::CreateTooltip(
+      "This is the type of units that your robot will be using."
+      "For example, if your flywheel will be using rotations, then use the "
```
prateekma, line 91: I think we should say something like "if you want your flywheel gains in terms of radians, use the radians unit."

prateekma (approved):

PR #9, opened 2021-02-14
Finish feedback gain unit conversions

### LINKED ISSUE #8: Convert timing for onboard velocity gains

Piphi5:
Fixes #8 

Divides by time conversion for native controller outputs. (e.g. SparkMax controller uses RPM -> divide by 60 to convert gains for that controller).

### DIFF
--- a/src/main/native/cpp/view/Analyzer.cpp
+++ b/src/main/native/cpp/view/Analyzer.cpp
@@ -351,6 +351,13 @@ void Analyzer::Display() {
       if (ImGui::Combo("Gain Preset", &m_selectedPreset, kPresetNames,
                        IM_ARRAYSIZE(kPresetNames))) {
         m_settings.preset = m_presets[kPresetNames[m_selectedPreset]];
+        if (m_settings.preset != presets::kWPILibNew &&
+            m_settings.preset != presets::kWPILibOld &&
+            m_settings.preset != presets::kDefault) {
+          m_settings.calcGains = true;
+        } else {
+          m_settings.calcGains = false;
+        }
         Calculate();
       }
       ImGui::SameLine();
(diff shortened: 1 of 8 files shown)


src/main/native/include/sysid/analysis/FeedbackControllerPreset.h
```diff
@@ -17,6 +17,8 @@ struct FeedbackControllerPreset {
   /** The conversion factor between volts and the final controller output. */
   double outputConversionFactor;
 
+  double outputVelocityTimeFactor;
```
prateekma, line 20: Can you add a comment above this explaining what it does? Maybe even clarify that the double above this one doesn't account for time.

Piphi5: Does:
`The conversion factor for using controller output for velocity gains. This is necessary as some controllers do velocity controls with different time units.`

clearly explain it?

prateekma: Yeah that sounds good.

src/main/native/include/sysid/analysis/AnalysisManager.h
```diff
@@ -212,6 +213,11 @@ class AnalysisManager {
   // controller preset, LQR parameters, acceleration window size, etc.
   const Settings& m_settings;
 
+  // Conversion factors. This contains values to convert feedback gains into.
+  int& m_epr;
+  double& m_gearing;
+  bool& m_calcGains;
```
prateekma, line 219: I would prefer if these were also in the `Settings` struct.

src/main/native/include/sysid/analysis/FeedbackAnalysis.h
```diff
@@ -37,7 +37,8 @@ struct LQRParameters {
  */
 std::tuple<double, double> CalculatePositionFeedbackGains(
     const FeedbackControllerPreset& preset, const LQRParameters& params,
-    const FeedforwardGains& feedforwardGains);
+    const FeedforwardGains& feedforwardGains, const double& gearing,
```
prateekma, line 40: double and int can be pass-by-value

src/main/native/include/sysid/analysis/FeedbackAnalysis.h
```diff
@@ -50,5 +51,6 @@ std::tuple<double, double> CalculatePositionFeedbackGains(
  */
 std::tuple<double, double> CalculateVelocityFeedbackGains(
     const FeedbackControllerPreset& preset, const LQRParameters& params,
-    const FeedforwardGains& feedforwardGains);
+    const FeedforwardGains& feedforwardGains, const double& gearing,
```
prateekma, line 54: double and int can be pass-by-value

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -421,6 +429,31 @@ void Analyzer::Display() {
...
+      sysid::CreateTooltip(
+          "This is the gearing between your encoder and your shaft");
+
+      ImGui::SetNextItemWidth(ImGui::GetFontSize() * 10);
+      if (ImGui::InputInt("EPR", &m_epr, 0) && m_epr > 0) {
+        Calculate();
+      }
+      sysid::CreateTooltip("This is the edges per rotation for your encoder");
+
+      if (ImGui::Checkbox("Convert Gains", &m_calcGains)) {
```
prateekma, line 450: I think we should have this at the top and only make the EPR and Gearing options show up if it's selected.

PR #15, opened 2021-02-15
Allow units override in Analyzer

prateekma:
This lets users change the units and units per rotation when analyzing
their data in case they messed up entering these values in the Logger.


Piphi5:
Does this just change the unit display or does it do conversions?

prateekma:
If you change the units per rotation, then it also does conversions. Similar to the Logger, if you select rotational units, it hardcodes the units per rotations at 1, 6.28 or 360.

PR #22, opened 2021-02-17
Limit size of data sent to ImPlot

### LINKED ISSUE #21: Arm Plotting Crashed
Voltage and Time domain plotting crashed the app with this error message:
```
sysid: /__w/thirdparty-imgui/thirdparty-imgui/imgui/imgui.cpp:4113: void AddDrawListToDrawData(ImVector<ImDrawList*>*, ImDrawList*): 
Assertion `draw_list->_VtxCurrentIdx < (1 << 16) && "Too many vertices in ImDrawList using 16-bit indices. Read comment above"' failed. 
Aborted (core dumped)
```

A possible cause could be that I left the test running for a relatively large amount of time.

[sysid_data20210217-0030.zip](https://github.com/wpilibsuite/sysid/files/5993736/sysid_data20210217-0030.zip)

Do note that it didn't crash for another, smaller json.

prateekma:
The max number of points that can be plotted in one plot is now 4096.
Closes #21

### DIFF
--- a/src/main/native/cpp/view/Analyzer.cpp
+++ b/src/main/native/cpp/view/Analyzer.cpp
@@ -27,11 +27,23 @@ struct VoltageDomainPlotData {
   AnalysisType type;
 };
 
+// This is the factor by which we should divide the data when plotting. For
+// example, if the size of the data is larger than 1 << 12, we will only visit
+// every other point.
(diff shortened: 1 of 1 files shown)


sciencewhiz:
Is the data at 200 hz? I'm not sure 20 seconds would be long enough for some of the drivetrain tests I've seen.

PR #38, opened 2021-02-27
Change format of JSON data

prateekma:
This splits up the JSON data into two separate formats -- one for
drivetrain and one for other general mechanisms. Furthermore, unecessary
data that was included in the old format is no longer included.


Piphi5:
Should we add a drivetrain angular test into the Simulation Tests?

PR #48, opened 2021-03-05
Make Logger NT selection async

prateekma:


prateekma:
Will wait for PR to split up the NT settings widget into libglassnt and use that instead.

PR #56, opened 2021-03-11
Add time domain simulation of feedforward gains

calcmogul:
Screenshot_2021-03-10_21-22-59


prateekma (approved):

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -253,7 +253,7 @@ void Analyzer::Display() {
   // Function that displays a read-only value.
   auto ShowGain = [](const char* text, double* data) {
     ImGui::SetNextItemWidth(ImGui::GetFontSize() * 4);
```
prateekma, line 255: We might need to increase this width a little to accommodate the extra significant figures.

prateekma (changes requested):

prateekma (approved):

PR #72, opened 2021-03-15
Fix bugs in AnalysisManager::TrimStepVoltageData

Piphi5:
Step Trimming in Analysis manager currently doesn't trim the data.

### DIFF
--- a/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -120,7 +120,7 @@ void TrimStepVoltageData(std::vector<PreparedData>* data) {
 
   for (size_t i = 0; i < data->size(); ++i) {
     // Get the current acceleration.
-    double acc = data->at(i).acceleration;
+    double acc = std::abs(data->at(i).acceleration);
 
     // If we are not in caution, the acceleration values are still increasing.
     if (!caution) {
@@ -146,6 +146,8 @@ void TrimStepVoltageData(std::vector<PreparedData>* data) {
       break;
     }
   }
+
+  data->erase(data->begin(), data->begin() + idx);
 }
 
 /**


calcmogul:
This improves the fit for Ka. Ground truth is Ks = 0, Kv = 1.98, Ka = 0.2.

```
main:
Ks 0.025672
Kv 1.971
Ka 0.18969
R^2 0.9935

PR #72:
Ks 0.013401
Kv 1.9755
Ka 0.19335
R^2 0.99671
```
Here's the JSON I used.
[integration_test_sysid_data20210307-1828.txt](https://github.com/wpilibsuite/sysid/files/6143516/integration_test_sysid_data20210307-1828.txt)

prateekma (dismissed):

calcmogul:
TrimStepVoltageData() doesn't seem to initialize the acceleration properly, so it trims the positive acceleration, but not the negative. The following change fixed that and improved the fit more.
```patch
diff --git a/src/main/native/cpp/analysis/AnalysisManager.cpp b/src/main/native/cpp/analysis/AnalysisManager.cpp
index 607e8f4..26b580e 100644
--- a/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -120,7 +120,7 @@ void TrimStepVoltageData(std::vector<PreparedData>* data) {
 
   for (size_t i = 0; i < data->size(); ++i) {
     // Get the current acceleration.
-    double acc = data->at(i).acceleration;
+    double acc = std::abs(data->at(i).acceleration);
 
     // If we are not in caution, the acceleration values are still increasing.
     if (!caution) {
```
```
Ks 0.000978
Kv 1.9798
Ka 0.19839
R^2 1
```

calcmogul:
By the way, both trimming fixes improves the new-ols fit considerably as well.
```
Ks -2.8707E-08
Kv 1.98
Ka 0.20003
R^2 1
```

prateekma (approved):

calcmogul (approved):

PR #82, opened 2021-03-26
Upload bad JSONs from integration tests

Piphi5:
If the integration test fails, it will store the json of the failed
test and for CI it will upload them as artifacts for debugging purposes.

~~TODO: Remove part that fails all of the integration tests~~

### DIFF
--- a/src/integrationtest/native/cpp/SimulationTest.cpp
+++ b/src/integrationtest/native/cpp/SimulationTest.cpp
@@ -114,17 +114,46 @@ class IntegrationTest : public ::testing::Test {
       auto ff = std::get<0>(output.ff);
       auto trackWidth = output.trackWidth;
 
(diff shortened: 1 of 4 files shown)


src/integrationtest/native/cpp/SimulationTest.cpp
```diff
@@ -114,6 +114,8 @@ class IntegrationTest : public ::testing::Test {
       auto ff = std::get<0>(output.ff);
       auto trackWidth = output.trackWidth;
 
+      wpi::outs() << "Kv: " << ff[1] << "\nKa: " << ff[2] << "\n";
+      wpi::outs().flush();
```
calcmogul, line 118: What about ff[0] and ff[3] for elevator and arm?

calcmogul (changes requested):

calcmogul (approved):

PR #83, opened 2021-03-26
Add input box for velocity denominator unit

prateekma:
@calcmogul Do you have any ideas for a better name? Basically, it represents what unit is used in the denominator for velocity measurements; i.e. 0.1 for CTRE and 60.0 for REV.


calcmogul:
"Velocity denominator units (s)"?

Piphi5:
Also, a tooltip could be useful?

PR #84, opened 2021-03-26
Move plots to separate view and add plot legend

Piphi5:
This should let users more easily see how changing settings like window size and motion threshold change their data.

### DIFF
--- a/src/main/native/cpp/view/Analyzer.cpp
+++ b/src/main/native/cpp/view/Analyzer.cpp
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <exception>
+#include <stdexcept>
 #include <thread>
 
 #include <glass/Context.h>
@@ -52,7 +53,7 @@ void Analyzer::Display() {
         m_manager = std::make_unique<AnalysisManager>(*m_location, m_settings,
                                                       m_logger);
         m_type = m_manager->GetAnalysisType();
-        Calculate();
+        RefreshInformation();
       } catch (const std::exception& e) {
         // If we run into an error here, let's just ignore it and make the user
         // explicitly select their file.
@@ -85,6 +86,7 @@ void Analyzer::Display() {
     if (ImGui::Combo("Dataset", &m_settings.dataset, AnalysisManager::kDatasets,
                      m_type == analysis::kDrivetrain ? 9 : 3)) {
       Calculate();
+      PrepareGraphs();
     }
     ImGui::SameLine(width - ImGui::CalcTextSize("Reset").x);
     if (ImGui::Button("Reset")) {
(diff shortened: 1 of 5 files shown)


src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -480,9 +454,27 @@ void Analyzer::SelectFile() {
   }
 }
 
+bool Analyzer::LoadPlots() {
+  // See if the plots are loaded
+  bool graphLoaded =
+      m_plot.DisplayVoltageDomainPlots() && m_plot.DisplayTimeDomainPlots();
```
prateekma, line 460: We can consolidate these two into one method in `AnalyzerPlot` now.

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -53,6 +53,7 @@ void Analyzer::Display() {
                                                       m_logger);
         m_type = m_manager->GetAnalysisType();
         Calculate();
+        PrepareData();
```
prateekma, line 56: Call this before `Calculate()`.

prateekma (changes requested):

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -231,31 +218,18 @@ void Analyzer::Display() {
...
 
-      // Show time domain diagnostic plots.
-      if (ImGui::BeginPopupModal("Time-Domain Diagnostics")) {
-        m_plot.DisplayTimeDomainPlots();
-        // Button to close popup.
-        if (ImGui::Button("Close")) {
-          ImGui::CloseCurrentPopup();
-        }
-        ImGui::EndPopup();
+      ImGui::BeginChild("Diagnostic Plots");
```
calcmogul, line 224: If we change this from `ImGui::BeginChild()` to `ImGui::Begin()`, we can put these in a separate window to make things less busy. Seems to work fine, and the only thing extra we need to do is set the window size.

calcmogul (changes requested):

src/main/native/cpp/view/Analyzer.cpp
Piphi5: I think calling it before `Calculate()` segfaults it because `PrepareData()` loads the graph data that requires feedforwards. What would be the best way of resolving this? Should we have a separate method solely to load graph data?

calcmogul: I think it crashed when doing it in the other order.

src/main/native/cpp/view/Analyzer.cpp
Piphi5: Would we move this outside of the Feedforward Dropdown so that the window would pop up at the beginning of the program?

src/main/native/cpp/view/Analyzer.cpp
prateekma: Yes I think we should have a separate function to load plot data.

src/main/native/cpp/view/Analyzer.cpp
prateekma: Whichever you think makes sense. I would prefer that, yes.

Piphi5:
Current setup:
image

prateekma (commented):
The one concern I have about this approach is that when we add a slider that will call `RefreshInformation()`, it will spawn a million threads, which will all be blocked until they can grab the lock. Effectively this means that the plots will be stuck in the "Loading..." state forever.

Piphi5:
Could we minimize this by increasing the interval that the slider can travel?

calcmogul:
We could also  wait for the slider to not change for a bit before we dispatch to the data generation.

Piphi5:
> We could also wait for the slider to stop changing for a bit before we dispatch to the data generation.

Canceling the previous update thread seems like the most robust way of doing it. How would that be implemented?

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -513,6 +492,33 @@ void Analyzer::Calculate() {
   }
 }
 
+void Analyzer::AbortDataPrep() {
+  if (m_dataThread.joinable()) {
+    m_abortDataPrep.store(true);
```
calcmogul, line 497:
```suggestion
    m_abortDataPrep = true;
```

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -513,6 +492,33 @@ void Analyzer::Calculate() {
...
+    m_dataThread.join();
+    m_abortDataPrep.store(false);
```
calcmogul, line 499:
```suggestion
    m_abortDataPrep = false;
```

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -513,6 +492,33 @@ void Analyzer::Calculate() {
   }
 }
 
+void Analyzer::AbortDataPrep() {
```
calcmogul, line 495: This function should be called in Analyzer's destructor.

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -68,12 +69,15 @@ AnalyzerPlot::AnalyzerPlot(wpi::Logger& logger) : m_logger(logger) {
...
 
 void AnalyzerPlot::SetData(const Storage& data, const std::vector<double>& ff,
-                           AnalysisType type) {
+                           AnalysisType type, std::atomic<bool>& abort) {
   std::scoped_lock lock(m_mutex);
   auto& [slow, fast] = data;
 
   // Clear all data vectors.
   for (auto it = m_data.begin(); it != m_data.end(); ++it) {
+    if (abort.load()) {
```
calcmogul, line 78:
```suggestion
    if (abort) {
```

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -108,6 +112,9 @@ void AnalyzerPlot::SetData(const Storage& data, const std::vector<double>& ff,
   // velocity-portion voltage graph.
   double t = slow[0].timestamp;
   for (size_t i = 0; i < slow.size(); i += sStep) {
+    if (abort.load()) {
```
calcmogul, line 115:
```suggestion
    if (abort) {
```

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -133,6 +140,9 @@ void AnalyzerPlot::SetData(const Storage& data, const std::vector<double>& ff,
   // acceleration-portion voltage graph.
   t = fast[0].timestamp;
   for (size_t i = 0; i < fast.size(); i += fStep) {
+    if (abort.load()) {
```
calcmogul, line 143:
```suggestion
    if (abort) {
```

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -278,4 +293,18 @@ void AnalyzerPlot::DisplayTimeDomainPlots() {
...
+  // See if the plots are loaded
+  bool graphLoaded = DisplayVoltageDomainPlots() && DisplayTimeDomainPlots();
+
+  // The plots have already been loaded if its currently loaded and it was
+  // previously loaded
+  bool alreadyLoaded = graphLoaded && m_prevGraphLoaded;
+
+  // Update previously loaded value
+  m_prevGraphLoaded = graphLoaded;
+  return alreadyLoaded;
```
calcmogul, line 309: Checking m_prevGraphLoaded is redundant since we only care about the current value, not whether it changed since the last value.

calcmogul (changes requested):

src/main/native/cpp/view/AnalyzerPlot.cpp
Piphi5: The issue is we have to set the scroll bar position just after the graphs have loaded.

PeterJohnson:
> Current setup:
> image

In this picture, the second graph on the right seems to have an error in the legend--shouldn't the blue be the collected data and the red the fit (the legend has the opposite).

src/main/native/cpp/view/AnalyzerPlot.cpp
calcmogul: Right, but this code checks when both variables return true, not when one of them goes from false to true.

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -84,7 +85,7 @@ void Analyzer::Display() {
     ImGui::SetNextItemWidth(ImGui::GetFontSize() * 15);
     if (ImGui::Combo("Dataset", &m_settings.dataset, AnalysisManager::kDatasets,
                      m_type == analysis::kDrivetrain ? 9 : 3)) {
-      Calculate();
+      Calculate();  // TODO: Determine if Plot Method needed here.
```
calcmogul, line 88: Why the TODO?

src/main/native/include/sysid/view/AnalyzerPlot.h
```diff
@@ -43,17 +44,29 @@ class AnalyzerPlot {
    * Sets the raw data to be displayed on the plots.
    */
   void SetData(const Storage& data, const std::vector<double>& ff,
-               AnalysisType type);
+               AnalysisType type, std::atomic<bool>& abort);
 
   /**
    * Displays voltage-domain plots.
+   *
+   * @return if plots aren't in the loading state
```
calcmogul, line 52: So the function never returns if the plots are in the loading state? Did you mean to say it returns true if the plots aren't in the loading state?

src/main/native/include/sysid/view/AnalyzerPlot.h
```diff
@@ -43,17 +44,29 @@ class AnalyzerPlot {
...
    */
-  void DisplayVoltageDomainPlots();
+  bool DisplayVoltageDomainPlots();
 
   /**
    * Displays time-domain plots.
+   *
+   * @return if plots aren't in the loading state
```
calcmogul, line 59: So the function never returns if the plots are in the loading state? Did you mean to say it returns true if the plots aren't in the loading state?

calcmogul (changes requested):

src/main/native/cpp/view/Analyzer.cpp
Piphi5: The plots didn't seem to be refreshed when the data sets are changed. I wasn't sure if there was a reason for this.

src/main/native/include/sysid/view/AnalyzerPlot.h
Piphi5: I meant for it to return true

src/main/native/cpp/view/AnalyzerPlot.cpp
Piphi5: Then should we keep the `m_prevgraphLoaded` as a member of the Analyzer?

Piphi5:
Would it be beneficial to have a button that users can press to reset the plots to the initial fitted axis?

prateekma:
No, they can just double click on the plot area and it will automatically do that

calcmogul:
The legends for the combined diagnostics should probably be put on the plot instead of to the side to make better use of horizontal space.
Screenshot_2021-04-07_22-21-59

calcmogul:
This is weird too.
Screenshot_2021-04-07_22-27-38

Piphi5:
So should we move all the legends into the plots? Or move the legends for all of the `CombinedDiagnostics` plots and only move the legends for the acceleration plots in the regular plotting window?

calcmogul:
I think we should move the legend onto all plot windows to maximize horizontal space. Voltage domain plots would use top-left and time domain plots would use top-right.

Piphi5:
is this better?
image

calcmogul:
Yep. We should do the same for the main plots too cuz they're so narrow now.

Piphi5:
image


> This is portion of the diagnostic plots window is weird too.

This should be fixed now.

calcmogul (approved):

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -370,6 +389,12 @@ void AnalyzerPlot::DisplayTimeDomainPlots(ImVec2 plotSize) {
       m_fitNextPlot[6] = false;
     }
   }
+  return true;
+}
+
+bool AnalyzerPlot::LoadPlots() {
+  // See if the plots are loaded
+  return DisplayVoltageDomainPlots() && DisplayTimeDomainPlots();
```
prateekma, line 397: Instead of calling these both at once, can we combine it all into one method? We don't need the ImVec2 parameter either.

prateekma (changes requested):

src/main/native/cpp/view/AnalyzerPlot.cpp
prateekma: Actually, let's keep these separate so there isn't a huge method. But we can remove the parameter.

Piphi5: Where are we removing the parameter from?

prateekma: The ImVec2 parameter to DisplayVoltageDomainPlots() and DisplayTimeDomainPlots().

Piphi5: How would we distinguish between the popup version and the regular version of the graphs?

prateekma: There is only one version now right?

calcmogul: There are two: this invocation with a default plot size and DisplayCombinedPlots()'s invocation with a non-default plot size.

Piphi5: Should we just do one method? Or do both work in this case since they do serve different purposes?

prateekma: Ah, I wasn't aware that we still wanted both methods. 👍

prateekma (approved):

PR #85, opened 2021-03-31
Add dt plots

Piphi5:
- Adds diagnostic plots to see the timing of data collection
- Stores test start positions for more robust identification of times
between tests

Example graph:
image

### DIFF
--- a/src/main/native/cpp/view/AnalyzerPlot.cpp
+++ b/src/main/native/cpp/view/AnalyzerPlot.cpp
@@ -28,7 +28,8 @@ static ImPlotPoint Getter(void* data, int idx) {
 
 template <typename Model>
 static std::vector<std::vector<ImPlotPoint>> PopulateTimeDomainSim(
-    const std::vector<PreparedData>& data, size_t step, Model model) {
+    const std::vector<PreparedData>& data,
+    const std::array<double, 4>& startTimes, size_t step, Model model) {
   // Create the vector of ImPlotPoints that will contain our simulated data.
   std::vector<std::vector<ImPlotPoint>> pts;
   std::vector<ImPlotPoint> tmp;
@@ -44,12 +45,15 @@ static std::vector<std::vector<ImPlotPoint>> PopulateTimeDomainSim(
     auto dt = units::second_t{now.timestamp} - units::second_t{pre.timestamp};
     t += dt;
 
-    // If there's a large gap or the time went backwards, it's a new
-    // section of data, so reset the model state
-    if (dt < 0_s || dt > 1_s) {
-      pts.emplace_back(std::move(tmp));
-      model.Reset(now.position, now.velocity);
-      continue;
+    // If the current time stamp and previous time stamp are across a test's
+    // start timestamp, it is the start of a new test and the model needs to be
+    // reset.
+    for (const auto& startTime : startTimes) {
+      if (now.timestamp >= startTime && pre.timestamp <= startTime) {
+        pts.emplace_back(std::move(tmp));
+        model.Reset(now.position, now.velocity);
+        continue;
+      }
     }
 
     model.Update(units::volt_t{pre.voltage}, dt);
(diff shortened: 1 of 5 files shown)


src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -127,6 +132,17 @@ void AnalyzerPlot::SetData(const Storage& data, const std::vector<double>& ff,
...
     m_data[kChartTitles[3]].emplace_back(slow[i].timestamp - t,
                                          slow[i].acceleration);
+
+    if (i > 0) {
+      // If the current timestamp is in the startTimes array, it is the
+      // beginning of a new test and the dt will be inflated. Therefore we skip
+      // it to exclude that dt and effectively reset dt calculations.
+      if (std::find(startTimes.begin(), startTimes.end(), slow[i].timestamp) ==
+          startTimes.end()) {
+        double dt = (slow[i].timestamp - slow[i - 1].timestamp) * 1000;
```
calcmogul, line 142: If `i` is the start time, subtracting the timestamp at `i - 1` will result in a large dt.

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -152,24 +168,36 @@ void AnalyzerPlot::SetData(const Storage& data, const std::vector<double>& ff,
                                          fast[i].velocity);
     m_data[kChartTitles[5]].emplace_back(fast[i].timestamp - t,
                                          fast[i].acceleration);
+    if (i > 0) {
+      // If the current timestamp is in the startTimes array, it is the
+      // beginning of a new test and the dt will be inflated. Therefore we skip
+      // it to exclude that dt and effectively reset dt calculations.
+      if (std::find(startTimes.begin(), startTimes.end(), fast[i].timestamp) ==
+          startTimes.end()) {
+        double dt = (fast[i].timestamp - fast[i - 1].timestamp) * 1000;
```
calcmogul, line 177: Same comments as earlier.

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -278,4 +306,22 @@ void AnalyzerPlot::DisplayTimeDomainPlots() {
       }
     }
   }
+  if (m_fitNextPlot[6]) {
+    ImPlot::FitNextPlotAxes();
+  }
```
calcmogul, line 311: We should clamp the y minimum to 0 since negative periods are impossible.

calcmogul (changes requested):

calcmogul:
It's not clear from the screenshot what the nomimal dt is supposed to be. We could draw a flat line at the mean perhaps.

src/main/native/cpp/view/AnalyzerPlot.cpp
Piphi5: Doesn't the if statement skip over values where the timestamp at `i` is a start time?

src/main/native/cpp/view/AnalyzerPlot.cpp
Piphi5: Do I set the plot limits or should I remove any negative datapoints?

Piphi5:
@calcmogul does this work?
image

prateekma:
Instead of autoscaling it, we should set the limits on this one because most dts will always be between 0 and 25.

Piphi5:
@prateekma 
is this sizing (0-50ms) good?
image

prateekma:
Yes

calcmogul (approved):

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -278,4 +327,27 @@ void AnalyzerPlot::DisplayTimeDomainPlots() {
       }
     }
   }
+  if (m_fitNextPlot[6]) {
+    ImPlot::SetNextPlotLimits(0, m_dtMaxTime, 0, 50);
```
prateekma, line 331: Can we add a fit call just for the x-axis after this? See the arguments here: https://github.com/epezent/implot/blob/master/implot.h#L522

src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -278,4 +327,28 @@ void AnalyzerPlot::DisplayTimeDomainPlots() {
...
+    ImPlot::FitNextPlotAxes(true, false, false, false);
```
Piphi5, line 332: @prateekma  is this what you were asking for?

prateekma: Yes

Piphi5: Is it good to merge then?

prateekma: Yes, I'll review again tonight to make sure but it should be good to go.

prateekma (approved):

PR #97, opened 2021-04-14
Update GUI Layout; move JSON converter to menu bar

### LINKED ISSUE #93: Move Frc-Char JSON converter to Menu Bar
This will leave more space for the other widgets.

Piphi5:
Updates GUI to make room for the generator and moves FRC-Char JSON to a popup that can be activated from the menu bar.
Fixes #93 
Overall Layout:
image

Converter Location:
image

Popup:
image

### DIFF
--- a/src/main/native/cpp/Main.cpp
+++ b/src/main/native/cpp/Main.cpp
@@ -10,6 +10,7 @@
 #include <glass/Window.h>
 #include <glass/WindowManager.h>
 #include <glass/other/Log.h>
+#include <imgui.h>
 #include <wpi/Logger.h>
 #include <wpi/Path.h>
 #include <wpi/StringRef.h>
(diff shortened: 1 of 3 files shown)


src/main/native/cpp/Main.cpp
```diff
@@ -2,6 +2,8 @@
 // Open Source Software; you can modify and/or share it under the terms of
 // the WPILib BSD license file in the root directory of this project.
 
+#include <imgui.h>
```
calcmogul, line 5: Why was this added? Should it be moved down with the other includes?

Piphi5: I'm not too sure as to why it was added. I think VSCode might've done it on its own?

calcmogul: Since we're using imgui calls directly, we should probably keep it but sort it with the other headers.

calcmogul (approved):

Issue #74, opened 2021-03-18
DARE exception prevents loading JSON and modifying settings to fix it


calcmogul:
1. Open the following JSON (extension changed to .txt because GitHub can't do .json):
[characterization-data20210317-1925_new.txt](https://github.com/wpilibsuite/sysid/files/6165503/characterization-data20210317-1925_new.txt)
2. Set the window size to 2 and the velocity threshold to 15 in the feedforward analysis settings

The DARE will throw an exception at discrete_algebraic_riccati_equation.cpp:37 saying that the system isn't stabilizable. This aborts the JSON load, which means the settings can't be fixed to avoid the error. Reopening sysid resets the settings to defaults, so the JSON can be loaded again.

We should report an error for the feedback analysis section, but still allow loading the JSON and tweaking the feedforward analysis settings.

PR #99, opened 2021-04-15
Refactor dt filtering and tighten integration test tolerances

calcmogul:
Time deltas in the range 0ms to 500ms are used to compute the mean time
delta. Then, data points are only considered that are within 1ms of that
mean.

### DIFF
--- a/src/integrationtest/native/cpp/SimulationTest.cpp
+++ b/src/integrationtest/native/cpp/SimulationTest.cpp
@@ -116,17 +116,24 @@ class IntegrationTest : public ::testing::Test {
       wpi::outs() << "Ks: " << ffGains[0] << "\n"
                   << "Kv: " << ffGains[1] << "\nKa: " << ffGains[2] << "\n";
       wpi::outs().flush();
-      EXPECT_NEAR(Kv, ffGains[1], 0.05);
-      EXPECT_NEAR(Ka, ffGains[2], 0.20);
+      if (m_settings.mechanism == sysid::analysis::kArm) {
+        // The acceleration fit (used only for arms) is worse than the next
+        // velocity fit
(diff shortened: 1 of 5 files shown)


Piphi5:
Wouldn't it be good (for development purposes at least) to be able to know/plot the values of all of the dt's? Just not sure how we would integrate the two types of dt's (maybe different colors for the ones removed) unless we want to just have the full data available for analysis through #100.

calcmogul:
If we plotted both raw and filtered dts, the main difference is values within 1ms of the mean would be a different color.

prateekma (approved):

Piphi5:
> Are the dts outside the 0ms to 500ms range what you had in mind?
Yeah, just so we know they exist.

prateekma (commented):
I wonder if using the median would be better here? Since the dts usually tend to have pretty high outliers, I'm worried that we might get a mean with no dts in a 1 ms neighborhood of the mean.

calcmogul:
The problem with that is the quality of the data rapidly deteriorates as dt deviates from the mean. There's a lot of outliers, but the mean is generally spot-on because there's a clump of datapoints close to the mean.

prateekma:
Ok, I agree that if it works in CI, it will probably work in all other situations. I am still thinking of a better error reporting pipeline (also useful for #74) so we can work that in later.

Piphi5:
Also, could we run the Integration Test CI multiple times to make sure that it really works?

prateekma:
I reran it three times and it passed always.

Piphi5:
So are all the dt's graphed or are negative ones excluded?

calcmogul:
Negative is excluded because their nonsensical.

Piphi5 (approved):

PR #100, opened 2021-04-15
Add JSON to CSV Converter

### LINKED ISSUE #91: Add csv download option to analyzer
This would make it easier to look at the raw data.

Piphi5:
Fixes #91 
Allows users to select a SysId JSON and have it be converted into a CSV file. This lets users better look at individual values and facilitates looking at the data in other programs / languages.

CSV files are saved with the following convention: `sysid-json-name(Test, Units).csv`
e.g:
[sysid_data20210320-0128(Simple, Meters).zip](https://github.com/wpilibsuite/sysid/files/6315595/sysid_data20210320-0128.Simple.Meters.zip)

### DIFF
--- a/src/main/native/cpp/Main.cpp
+++ b/src/main/native/cpp/Main.cpp
@@ -132,9 +132,13 @@ int main() {
     }
 
     bool frcCharConvert = false;
+    bool toCSV = false;
     if (ImGui::BeginMenu("JSON Converters")) {
       if (ImGui::MenuItem("FRC-Char Converter")) {
         frcCharConvert = true;
+      }
+      if (ImGui::MenuItem("JSON to CSV Converter")) {
+        toCSV = true;
       }
 
       ImGui::EndMenu();
@@ -147,8 +151,21 @@ int main() {
       frcCharConvert = false;
     }
 
+    if (toCSV) {
+      ImGui::OpenPopup("SysId JSON to CSV Converter");
+      toCSV = false;
+    }
+
     if (ImGui::BeginPopupModal("FRC-Char Converter")) {
-      gJSONConverter->Display();
+      gJSONConverter->DisplayFRCCharConvert();
+      if (ImGui::Button("Close")) {
+        ImGui::CloseCurrentPopup();
+      }
+      ImGui::EndPopup();
+    }
+
+    if (ImGui::BeginPopupModal("SysId JSON to CSV Converter")) {
+      gJSONConverter->DisplayCSVConvert();
       if (ImGui::Button("Close")) {
         ImGui::CloseCurrentPopup();
       }
--- a/src/main/native/cpp/analysis/JSONConverter.cpp
+++ b/src/main/native/cpp/analysis/JSONConverter.cpp
@@ -14,6 +14,7 @@
 #include <wpi/raw_istream.h>
 #include <wpi/raw_ostream.h>
 
+#include "sysid/Util.h"
 #include "sysid/analysis/AnalysisManager.h"
 #include "sysid/analysis/AnalysisType.h"
 
@@ -30,7 +31,7 @@ static constexpr size_t kRPosCol = 6;
 static constexpr size_t kLVelCol = 7;
 static constexpr size_t kRVelCol = 8;
 
-std::string sysid::ConvertJSON(wpi::StringRef path, wpi::Logger& logger) {
+static wpi::json GetJSON(wpi::StringRef path, wpi::Logger& logger) {
   std::error_code ec;
   wpi::raw_fd_istream input{path, ec};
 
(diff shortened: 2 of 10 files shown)


src/main/native/cpp/Main.cpp
```diff
@@ -2,6 +2,8 @@
 // Open Source Software; you can modify and/or share it under the terms of
 // the WPILib BSD license file in the root directory of this project.
 
+#include <functional>
```
calcmogul, line 5: vs code strikes again?

src/main/native/cpp/analysis/JSONConverter.cpp
```diff
@@ -41,6 +42,11 @@ std::string sysid::ConvertJSON(wpi::StringRef path, wpi::Logger& logger) {
   wpi::json ojson;
```
prateekma, line 42: Rename this to just `json`. `ojson` stood for "old json" but that doesn't make sense in this method.

src/main/native/cpp/analysis/JSONConverter.cpp
```diff
@@ -84,10 +90,67 @@ std::string sysid::ConvertJSON(wpi::StringRef path, wpi::Logger& logger) {
...
+  wpi::json json = GetJSON(path, logger);
+
+  auto type = sysid::analysis::FromName(json.at("test").get<std::string>());
+  auto factor = json.at("unitsPerRotation").get<double>();
+  auto unit = json.at("units").get<std::string>();
+
+  // Naming: {sysid-json-name}(Test, Units).csv
+  std::string loc = path.rsplit(".json").first.str() + "(" + type.name + ", " +
+                    unit + ").csv";
+  std::ofstream outputFile{loc};
```
prateekma, line 112: Use `wpi::raw_fd_ostream` and remove corresponding `fstream` include.

prateekma (changes requested):

src/main/native/cpp/Main.cpp
Piphi5: ah yep

prateekma:
Can you rebase on main?

src/main/native/cpp/analysis/JSONConverter.cpp
```diff
@@ -84,10 +89,68 @@ std::string sysid::ConvertJSON(wpi::StringRef path, wpi::Logger& logger) {
...
+                    unit + ").csv";
+  wpi::raw_fd_ostream outputFile{loc, ec};
+
+  if (ec) {
+    throw std::runtime_error("Unable to write to: " + loc);
+  }
+
+  outputFile << "Timestamp, Test, ";
+  if (type == analysis::kDrivetrain || type == analysis::kDrivetrainAngular) {
+    outputFile << "Left Volts, Right Volts, Left Position, Right Position, "
```
calcmogul, line 120: This adds invisible spaces to the CSV fields. Generally, the CSV format has these stripped.

calcmogul (changes requested):

src/main/native/cpp/view/JSONConverter.cpp
```diff
@@ -54,3 +55,11 @@ void JSONConverter::Display() {
     ImGui::EndPopup();
   }
 }
+
+void JSONConverter::DisplayFRCCharConvert() {
+  DisplayConverter("Select FRC-Char JSOn", sysid::ConvertJSON);
```
prateekma, line 60: typo

prateekma (changes requested):

src/main/native/cpp/analysis/JSONConverter.cpp
Piphi5: is there a way to get the unit abbreviation or should I just do a bunch of ifs?

calcmogul: If we use the units library, we can grab the string suffix programmatically. That's not how the code's currently set up though.

Piphi5: or could we just append the entire unit name?

calcmogul: We could, but engineering convention for plot axis labels uses the name of the quantity followed by the unit abbreviation in parentheses.

Piphi5: So if statements would be the way to go?

calcmogul: Yea, that would be easiest.

Piphi5:
Should we slap on the abbreviations to the normal graphs as well?

src/main/native/include/sysid/Util.h
```diff
@@ -25,4 +25,5 @@
 namespace sysid {
 void CreateTooltip(const char* text);
 std::vector<std::string> Split(const std::string& s, char c);
+std::string GetAbbreviation(std::string unit);
```
prateekma, line 28: Use const-ref or wpi::StringRef

calcmogul:
> Should we slap on the abbreviations to the normal graphs as well?

Yes, if they aren't there already.

calcmogul (approved):
SysId plot labels and the plots I made from the CSV data look good.

prateekma (approved):

PR #102, opened 2021-04-16
Fix dt plot y-axis autofit

Piphi5:


calcmogul (approved):
That fixed it for me.

PR #24, opened 2021-02-19
Add initial project generation infrastructure

Piphi5:
TODO:
Fixes #14 
- [x] Finalize Generator GUI
- [x] Finish Single-Sided Motor Project
- [x] Finish Drive Project
- [x] Create Tests to ensure the validity of these projects
- [x] Fix #37 By adding Romi Preset
- [x] Update data to match #38 
- [x] Add the proper CPR conversions for motor controllers
- [x] Fix #41 by adding SysId Classes
- [x] Fix widget sizing
- [x] ~~Set CAN update rates to 5 ms (need to figure out CAN issues)~~ (File an issue, and resolve later so that we can get the fundamental generator merged + tested sooner)
- [x] Have settings for timing and average for different types of encoders
- [x] ~~Add venom support~~ (File an issue, and resolve later so that we can get the fundamental generator merged + tested sooner)
- [x] Optimize the logging by auto calculating the voltage
- [x] Update Telemetry Docs
- [x] Fix linter errors
- [x] Update Usage Docs in README

PR #103, opened 2021-04-16
Replace std::for_each() with range-based for loop

calcmogul:


integration_test_project/src/main/cpp/Robot.cpp
```diff
@@ -49,8 +49,9 @@ class Robot : public frc::TimedRobot {
       wpi::outs() << "Collected " << m_counter << " data points.\n";
 
       std::stringstream ss;
-      std::for_each(m_data.begin(), m_data.end(),
-                    [&ss](auto& pt) { ss << std::to_string(pt) << ", "; });
+      for (const auto& pt : m_data) {
+        ss << std::to_string(pt) << ", ";
+      }
```
Piphi5, line 54: Should I update the Logger Classes in #24 to match this?

calcmogul: Yea, it's cleaner.

prateekma (approved):

PR #105, opened 2021-04-17
Add RMSE metric to assess feedforward accuracy

Piphi5:
Shows the Root Mean Squared Error of the Simulation Predictions vs Raw Data. Serves as a quantitative way of assessing SysId performance.
image

### DIFF
--- a/src/main/native/cpp/view/Analyzer.cpp
+++ b/src/main/native/cpp/view/Analyzer.cpp
@@ -199,6 +199,11 @@ void Analyzer::Display() {
       }
 
       ShowGain("r-squared", &m_rSquared);
+      ShowGain("Sim RMSE", m_plot.GetRMSE());
+      CreateTooltip(
+          "The Root Mean Squared Error (RMSE) of the simulation "
+          "predictions compared to the recorded data. It is essentially the "
+          "error of the simulated model in the recorded units.");
 
       double endY = ImGui::GetCursorPosY();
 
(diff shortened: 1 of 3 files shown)


src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -199,6 +199,7 @@ void Analyzer::Display() {
       }
 
       ShowGain("r-squared", &m_rSquared);
+      ShowGain("Sim RMSE", m_plot.GetRMSE());
```
prateekma, line 202: Can we add a tooltip explaining what this is? I don't expect everyone to be familiar with this acronym.

prateekma (changes requested):

src/main/native/cpp/view/Analyzer.cpp
Piphi5: Let me know if the tooltip is descriptive enough

src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -199,6 +199,11 @@ void Analyzer::Display() {
...
+      CreateTooltip(
+          "This is the Root Mean Squared Error (RMSE) of the Simulation "
```
prateekma, line 204: Remove "This is" (capitalize The) and no need to capitalize the s in simulation.

prateekma (changes requested):
One small change and should be good to go.

prateekma (approved):

PR #106, opened 2021-04-19
Shrink dt range for valid data

calcmogul:


src/integrationtest/native/cpp/SimulationTest.cpp
```diff
@@ -119,8 +119,8 @@ class IntegrationTest : public ::testing::Test {
       if (m_settings.mechanism == sysid::analysis::kArm) {
         // The acceleration fit (used only for arms) is worse than the next
         // velocity fit
-        EXPECT_NEAR(Kv, ffGains[1], 0.015);
-        EXPECT_NEAR(Ka, ffGains[2], 0.04);
+        EXPECT_NEAR(Kv, ffGains[1], 0.01);
+        EXPECT_NEAR(Ka, ffGains[2], 0.01);
```
Piphi5, line 123: How many tests have you done with these values?

calcmogul: Two, including this one.

Piphi5: Looks like we need to increase kA?

calcmogul: It's off by like 0.07. :/

Piphi5: Mac is cursed

calcmogul:
Locally, using the 1ms dt band makes it more accurate. It seems to be a tradeoff between noise magnitude and datapoint density.

PR #113, opened 2021-04-26
Filter dt as part of prepared data pipeline

calcmogul:
This required refactoring ComputeAcceleration() and CalculateCosine()
into one function that turns the raw data into prepared data.


calcmogul:
This refactor has a long way to go, but this PR did fix the fit for some flywheel data.

main:
Screenshot_2021-04-25_17-46-17

this PR:
Screenshot_2021-04-25_17-48-02

prateekma (approved):

PR #121, opened 2021-04-28
Convert Raw Data to Prepared Data after reading it

Piphi5:
Converts the raw data into the PreparedData struct once the data is
loaded. This reduces the number of template functions and makes the data
pipeline cleaner.

- [x] Fix the sim graphs getting messed up for some of the data.
- [x] Mac elevator fails

### DIFF
--- a/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -33,49 +33,6 @@
 using namespace sysid;
 
 /**
- * Helper method that applies a function onto a dataset.
- *
- * @tparam T The type of data that is stored within a StringMap
- *
- * @param data   A StringMap representing a specific dataset
- * @param action A void function that takes a StringRef representing a StringMap
- *               key and performs an action to the dataset stored with that
- *               passed key (e.g. applying a median filter)
- */
-template <typename T>
-void ApplyToData(const wpi::StringMap<T>& data,
-                 std::function<void(wpi::StringRef)> action) {
-  for (const auto& it : data) {
-    action(it.first());
-  }
-}
-
-/**
- * Helper method that applies a function onto a dataset.
- *
- * @tparam T The type of data that is stored within a StringMap
- *
- * @param data      A StringMap representing a specific dataset
- * @param action    A void function that takes a StringRef representing a
- *                  StringMap key and performs an action to the dataset stored
- *                  with that passed key (e.g. applying a median filter)
- * @param specifier A boolean function that takes a StringRef representing a
- *                  StringMap key and returns true if `action` should be run on
- *                  the dataset stored with that key
- */
-template <typename T>
-void ApplyToData(const wpi::StringMap<T>& data,
-                 std::function<void(wpi::StringRef)> action,
-                 std::function<bool(wpi::StringRef)> specifier) {
-  for (const auto& it : data) {
-    auto key = it.first();
-    if (specifier(key)) {
-      action(key);
-    }
-  }
-}
-
-/**
  * Converts a raw data vector into a PreparedData vector with only the
  * timestamp, voltage, position, and velocity fields filled out.
  *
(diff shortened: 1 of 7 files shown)


src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -91,3 +91,44 @@ units::second_t sysid::GetMeanTimeDelta(const Storage& data) {
...
+  // Starting points with the first value calculated by the median filter (since
+  // accel should be zero)
+  for (int i = 0; i < step; i++) {
+    data->at(i).velocity = data->at(step).velocity;
+  }
+
+  // Do the same procedure for the end datapoints
+  for (int i = data->size() - step; i < data->size(); i++) {
+    data->at(i).velocity = data->at(data->size() - step - 1).velocity;
+  }
```
calcmogul, line 124: This would be more accurate if you include the padded values in the actual filter calculation. The boundary condition is that the acceleration is zero at the first point, not that the acceleration is zero for half the window period.

src/main/native/include/sysid/analysis/FilteringUtils.h
```diff
@@ -98,7 +74,13 @@ units::second_t TrimStepVoltageData(std::vector<PreparedData>* data,
...
 /**
  * Compute the mean time delta of the given data.
+ *
+ * @param data A reference to all of the collected PreparedData
+ *
+ * @return The mean time delta for all the data points
  */
 units::second_t GetMeanTimeDelta(const Storage& data);
 
+void FilterAccelData(std::vector<PreparedData>* data);
```
calcmogul, line 84: This needs a Doxygen comment explaining what "filter accel" means in this context.

src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -91,3 +91,44 @@ units::second_t sysid::GetMeanTimeDelta(const Storage& data) {
...
+                             [motionThreshold](const auto& pt) {
+                               return std::abs(pt.voltage) <= 0 ||
+                                      std::abs(pt.velocity) < motionThreshold;
+                             }),
+              data->end());
+}
+
+void sysid::ApplyMedianFilter(std::vector<PreparedData>* data, int window) {
+  size_t step = window / 2;
+  frc::MedianFilter<double> medianFilter(window);
```
calcmogul, line 107: Since the MedianFilter is used in a .cpp file, add `#include <frc/MedianFilter.h>` to the top and remove it from FilteringUtils.h.

src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -86,114 +119,56 @@ static std::vector<PreparedData> Concatenate(
     dest.insert(dest.end(), ptr->cbegin(), ptr->cend());
   }
 
+  std::sort(dest.begin(), dest.end(), [](PreparedData a, PreparedData b) {
```
calcmogul, line 122:
```suggestion
  std::sort(dest.begin(), dest.end(), [](const auto& a, const auto& b) {
```

src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -213,13 +188,127 @@ static void TrimByTimeDelta(std::vector<PreparedData>* data,
...
+ * calculates acceleration and cosine (arm only) data, and trims the dynamic
+ * tests.
+ *
+ * @param data A pointer to a data vector recently created by the
+ * ConvertToPrepared method
+ * @param settings A reference to the analysis settings
+ * @param minStepTime A reference to the minimum dynamic test duration
+ * @param maxStepTime A reference to the maximum dynamic test duration
+ * @param unit The angular unit that the arm test is in (only for calculating
+ * cosine data)
```
calcmogul, line 230:
```suggestion
 * @param data A pointer to a data vector recently created by the
 *             ConvertToPrepared method
 * @param settings A reference to the analysis settings
 * @param minStepTime A reference to the minimum dynamic test duration
 * @param maxStepTime A reference to the maximum dynamic test duration
 * @param unit The angular unit that the arm test is in (only for calculating
 *             cosine data)
```

calcmogul (changes requested):

src/main/native/cpp/analysis/AnalysisManager.cpp
calcmogul, line 122: There should also be a comment here explaining why we're sorting.

src/main/native/cpp/analysis/FilteringUtils.cpp
Piphi5: So basically continue running the median filter by running Calculate with either the first or last value?

calcmogul: Yea.

Piphi5: Is the updated implementation what you were going for?

src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -91,3 +138,168 @@ units::second_t sysid::GetMeanTimeDelta(const Storage& data) {
...
+
+  // Load the median filter with the first value, step times for accurate
+  // behaviour at the start.
+
+  for (int i = 0; i < step; i++) {
+    medianFilter.Calculate(data->at(i).velocity);
```
calcmogul, line 160: Should this use `at(0)`, not `at(i)`? Only the first value should be replicated. Otherwise, you end up going backward since the next for loop starts at 0.

calcmogul (changes requested):

src/main/native/cpp/analysis/FilteringUtils.cpp
calcmogul: Not quite. See my latest comment.

src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -91,3 +138,168 @@ units::second_t sysid::GetMeanTimeDelta(const Storage& data) {
...
+              data->end());
+}
+
+void sysid::ApplyMedianFilter(std::vector<PreparedData>* data, int window) {
+  size_t step = window / 2;
+  frc::MedianFilter<double> medianFilter(window);
+
+  // Load the median filter with the first value, step times for accurate
+  // behaviour at the start.
+
```
calcmogul, line 158:
```suggestion
  // Load the median filter with the first value "step" number of times for
  // accurate initial behavior.
```
Also remove the blank line before the for loop since this comment refers to that block of code specifically.

src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -91,3 +138,168 @@ units::second_t sysid::GetMeanTimeDelta(const Storage& data) {
...
+  }
+
+  for (size_t i = 0; i < data->size(); i++) {
+    double median = medianFilter.Calculate(data->at(i).velocity);
+    if (i >= step) {
+      data->at(i - step).velocity = median;
+    }
+  }
+
+  // Run the median filter for the last values and fill it with the last value
```
calcmogul, line 170: Fill what with the last value? The median filter or the filtered data?

calcmogul (changes requested):

src/main/native/cpp/analysis/FilteringUtils.cpp
Piphi5: Is this message clearer?

calcmogul (approved):

calcmogul:
Increasing the dt passband, or removing it altogether, didn't seem to help the fit on that arm test that failed; Kv got better but Ka got significantly worse. That means it's probably not caused by a lack of data.

Piphi5:
In the case of the arm it could be because of the fact that we're still using the old algorithm. I found that the failed Elevator and Flywheels from previous pushes removed too much data.

prateekma (approved):

PR #127, opened 2021-05-01
Use higher order central finite difference for acceleration calc

calcmogul:
This gives much less noisy acceleration estimates.

### DIFF
--- a/src/integrationtest_analysis/native/cpp/AnalysisTest.cpp
+++ b/src/integrationtest_analysis/native/cpp/AnalysisTest.cpp
@@ -110,15 +110,8 @@ class AnalysisTest : public ::testing::Test {
       wpi::outs() << "Ks: " << ffGains[0] << "\n"
                   << "Kv: " << ffGains[1] << "\nKa: " << ffGains[2] << "\n";
       wpi::outs().flush();
-      if (m_settings.mechanism == sysid::analysis::kArm) {
-        // The acceleration fit (used only for arms) is worse than the next
-        // velocity fit
-        EXPECT_NEAR(Kv, ffGains[1], 0.015);
-        EXPECT_NEAR(Ka, ffGains[2], 0.04);
-      } else {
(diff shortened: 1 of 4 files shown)


calcmogul:
Before:
before-central-finite-diff
After:
after-central-finite-diff

[arm-ci-failure.json.txt](https://github.com/wpilibsuite/sysid/files/6410278/arm-ci-failure.json.txt)

Piphi5:
Would it be worth it to add tests?

calcmogul:
Yea.

calcmogul:
Unit tests for central finite difference have been added.

calcmogul:
Removing the dt filtering improved the fit for the failed data I downloaded from previous CI runs for this PR.

Piphi5:
Wasn't the removal of those dts necessary for some data recorded from frc-char?

calcmogul:
It worked fine on the data I have locally, except for the following two. Their dts are also pretty good, so dt removal wouldn't have helped.

[top-characterization-data-20210317-1525_new.json.txt](https://github.com/wpilibsuite/sysid/files/6412083/top-characterization-data-20210317-1525_new.json.txt)
[bottom-characterization-data20210319-1504_new.json.txt](https://github.com/wpilibsuite/sysid/files/6412084/bottom-characterization-data20210319-1504_new.json.txt)

Piphi5:
What was #113  referring to then?

calcmogul:
The instability that PR fixed was probably caused by the poor acceleration calculation.

Piphi5:
But doesn't flywheel data use new ols that isn't dependent on accel?

calcmogul:
The acceleration affects the data trimming. I just tried running main without the dt removal, and it's still stable. I'm not sure why. The fact remains though that the CI was failing more often until I just removed it.

Piphi5:
Alright that makes sense

prateekma (approved):

PR #129, opened 2021-05-05
Create Integration Test Utility Functions

### LINKED ISSUE #108: Integration Test Projects that don't build don't terminate
As seen [here](https://github.com/Piphi5/sysid/runs/2385236947?check_suite_focus=true), whenever the integration test project fails to build, the test waits indefinitely to connect to the robot. Either terminating the test whenever the status for `std::system` != 0 or making the connect wait have a time out would fix this.

Piphi5:
Resolves #108 

Turns out that asserting during a Test Suite Setup will not exit the test so std::exit(1) is called instead. Faulty code now fails [immediately ](https://github.com/Piphi5/sysid/actions/runs/812253950) rather than [maxing out ](https://github.com/Piphi5/sysid/actions/runs/757774522) the 6 hr CI time limit.

Also moves Launching Sim, Connecting to NT, and Killing NT to utility functions to reduce code duplication.


src/integrationtest_utils/native/include/IntegrationUtils.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Connects to a Simulation NT Entry and configures the kill variable to not
+ * prematurely kill the simulated code. This assumes that the simulated code
+ * uses a NT kill entry to end the simulation.
+ *
+ * @param m_nt the Network Tables Intstance
+ * @param m_kill the Network Tables Entry responsible for killing the simulated
+ * code.
+ */
+void Connect(NT_Inst m_nt, NT_Entry m_kill);
```
calcmogul, line 32: Remove the `m_` prefix. That prefix is only for private member variables.

src/integrationtest_utils/native/include/IntegrationUtils.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Connects to a Simulation NT Entry and configures the kill variable to kill
+ * the robot code. This assumes that the simulated code uses a NT kill entry to
+ * end the simulation.
+ *
+ * @param m_nt the Network Tables Intstance
+ * @param m_kill the Network Tables Entry responsible for killing the simulated
+ * code.
+ */
+void KillNT(NT_Inst m_nt, NT_Entry m_kill);
```
calcmogul, line 43: Remove the `m_` prefix. That prefix is only for private member variables.

calcmogul (changes requested):

src/integrationtest_utils/native/cpp/IntegrationUtils.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "IntegrationUtils.h"
+
+#include <wpi/SmallString.h>
+#include <wpi/SmallVector.h>
+#include <wpi/raw_ostream.h>
+#include <wpi/timestamp.h>
+
+#include "gtest/gtest.h"
+
+void LaunchSim(std::string projectDirectory) {
```
calcmogul, line 14: This argument is unused.

calcmogul (changes requested):

src/integrationtest_utils/native/cpp/IntegrationUtils.cpp
Piphi5: Good catch, that would've been bad.

calcmogul:
Why are the two LaunchSim() calls different?

Piphi5:
One calls the integration test project, the other needs to call either the General Mechanism or Drivetrain projects.

calcmogul (approved):

build.gradle
```diff
@@ -162,9 +162,12 @@ model {
     }
     sysidAnalysisIntegrationTest(GoogleTestTestSuiteSpec) {
       testing $.components.sysid
-      sources.cpp.source {
-        srcDirs "src/integrationtest_analysis/native/cpp"
-        include "**/*.cpp"
+      sources.cpp {
+        source {
+        srcDirs "src/integrationtest_analysis/native/cpp", "src/integrationtest_utils/native/cpp"
+        include "**/*.cpp", "**/*.h"
```
prateekma, line 168: Indent

build.gradle
```diff
@@ -176,9 +179,12 @@ model {
     }
     sysidGenerationIntegrationTest(GoogleTestTestSuiteSpec) {
       testing $.components.sysid
-      sources.cpp.source {
-        srcDirs "src/integrationtest_generation/native/cpp"
-        include "**/*.cpp"
+      sources.cpp {
+        source {
+        srcDirs "src/integrationtest_generation/native/cpp", "src/integrationtest_utils/native/cpp"
+        include "**/*.cpp", "**/*.h"
```
prateekma, line 185: Indent

prateekma (changes requested):

calcmogul (approved):

prateekma (approved):

PR #133, opened 2021-05-09
Fix filtered data lag in plots

### LINKED ISSUE #132: Median Filter Time offset for certain datasets
For a particularly noisy dataset, the median filter seems to have gotten the timestamps shifted.

image

Piphi5:
Resolves #132 

Uses Timestamp instead of dts to plot the data.

Also fixes some additional bugs in the current analysis pipeline.

image

### DIFF
--- a/src/main/native/cpp/analysis/FilteringUtils.cpp
+++ b/src/main/native/cpp/analysis/FilteringUtils.cpp
@@ -193,6 +193,19 @@ void sysid::FilterAccelData(std::vector<PreparedData>* data) {
 }
 
 /**
+ * Removes a substring from a string reference
+ *
+ * @param str The wpi::StringRef that needs modification
+ * @param removeStr The substring that needs to be removed
+ *
+ * @return an std::string without the specified substring
+ */
+static std::string RemoveStr(wpi::StringRef str, wpi::StringRef removeStr) {
+  auto pair = str.split(removeStr);
+  return pair.first.str() + pair.second.str();
+}
+
+/**
  * Figures out the max duration of the Dynamic tests
  *
  * @tparam S The size of the arrays in the raw data vector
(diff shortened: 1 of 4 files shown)


src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -10,7 +10,11 @@
 #include <frc/LinearFilter.h>
 #include <frc/MedianFilter.h>
 #include <units/math.h>
+#include <wpi/StringRef.h>
 #include <wpi/math>
+#include <wpi/raw_ostream.h>
```
calcmogul, line 15: Can this include be removed?

src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -10,7 +10,11 @@
...
+
+#include "sysid/analysis/Storage.h"
```
calcmogul, line 17: This include shouldn't be needed because it's already included by FilteringUtils.h.

calcmogul (changes requested):

calcmogul:
The quasistatic sim with that data misbehaves a bit. Is the higher velocity threshold what's causing it? Perhaps the low velocity threshold is hiding a bug.
Screenshot_2021-05-08_19-59-45

src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -10,7 +10,11 @@
 #include <frc/LinearFilter.h>
 #include <frc/MedianFilter.h>
 #include <units/math.h>
+#include <wpi/StringRef.h>
```
calcmogul, line 13: StringRef is used in FilteringUtils.h as well, so this include should be moved there.

Piphi5:
> The quasistatic sim with that data misbehaves a bit. Is the higher velocity threshold what's causing it? Perhaps the low velocity threshold is hiding a bug.

It seems like the kV gains are pretty bad with the low motion threshold. But the lower velocity threshold could be causing some issues as well

calcmogul (approved):

PR #135, opened 2021-05-11
Add time plotting python script

Piphi5:
This adds a time plotting python script that plots data over time. This is mainly intended to help troubleshoot a scenario in which a dataset breaks sysid. It supports both json and csv plotting.

Example plots:
General Mech:
image

Drivetrain:
image

### DIFF
--- a/scripts/time_plots.py
+++ b/scripts/time_plots.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+import json
+import matplotlib.pyplot as plt
+import pandas as pd
+import sys
+
+# Load data
+filename = sys.argv[1]
+
+# Make DataFrame to facillitate plotting
+if filename.endswith(".json"):
+    with open(filename) as json_file:
+        raw_data = json.load(json_file)
+    unit = raw_data["units"]
+
+    # Get Unit
+    if unit == "Meters":
+        abbreviation = "m"
+    elif unit == "Feet":
+        abbreviation = "ft"
+    elif unit == "Inches":
+        abbreviation = "in"
+    elif unit == "Degrees":
+        abbreviation = "deg"
+    elif unit == "Rotations":
+        abbreviation = "rot"
+    elif unit == "Radians":
+        abbreviation = "rad"
+    else:
+        raise ValueError("Invalid Unit")
+
+    # Make Columns
+    columns = ["Timestamp (s)", "Test"]
+    if "Drive" in raw_data["test"]:
+        columns.extend([
+            "Left Volts (V)",
+            "Right Volts (V)",
+            f"Left Position ({abbreviation})",
+            f"Right Position ({abbreviation})",
+            f"Left Velocity ({abbreviation}/s)",
+            f"Right Velocity ({abbreviation}/s)",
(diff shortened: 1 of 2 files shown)


scripts/time_plots.py
```diff
@@ -0,0 +1,80 @@
...
+        columns.extend([
+            "Volts (V)", f"Position ({abbreviation})",
+            f"Velocity ({abbreviation}/s)"
+        ])
+        unit_columns = columns[3:]
+
+    df = pd.DataFrame(columns=columns)
+    for test in (test for test in data.keys() if "-" in test):
+        df_data = [[pt[0]] + [test] + pt[1:] for pt in data[test]]
+        df = df.append(pd.DataFrame(df_data, columns=columns))
```
calcmogul, line 53: Isn't assigning to `df` redundant here?
```
>>> df=["a"]
>>> df.append("b")
>>> df
['a', 'b']
```

scripts/time_plots.py
```diff
@@ -0,0 +1,80 @@
...
+        ])
+        unit_columns = columns[4:8]
+    else:
+        columns.extend([
+            "Volts (V)", f"Position ({abbreviation})",
+            f"Velocity ({abbreviation}/s)"
+        ])
+        unit_columns = columns[3:]
+
+    df = pd.DataFrame(columns=columns)
```
calcmogul, line 50: `df` could use a more descriptive name.

scripts/time_plots.py
```diff
@@ -0,0 +1,80 @@
+import pandas as pd
```
calcmogul, line 1: Sorta these alphabetically if possible.

calcmogul (changes requested):

scripts/time_plots.py
Piphi5: by default the append method returns a copy, but I can set the in_place param to True if that would reduce the redundancy

prateekma:
@calcmogul Can you review again?

scripts/time_plots.py
```diff
@@ -0,0 +1,81 @@
+import json
```
calcmogul, line 1: Put `#!/usr/bin/env python3` on the first line of the file, then a blank line. Also, mark time_plots.py as executable by running `git update-index --chmod=+x time_plots.py`.

scripts/time_plots.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Get Unit
+    if unit == "Meters":
+        abbreviation = "m"
+    elif unit == "Feet":
+        abbreviation = "ft"
+    elif unit == "Inches":
+        abbreviation = "in"
+    elif unit == "Degrees":
+        abbreviation = "deg"
```
calcmogul, line 23: "Radians" should be added as a valid unit with the abbreviation "rad".

calcmogul (changes requested):

calcmogul:
Are you sure you committed the change to the execute bit?

calcmogul (approved):

Piphi5:
Should be updated now

PR #139, opened 2021-05-13
Include cstddef in OLS.h

Piphi5:
For some reason, linting started to fail due to an import not being included in the ols header file. This fixes it.

### DIFF
--- a/src/main/native/include/sysid/analysis/OLS.h
+++ b/src/main/native/include/sysid/analysis/OLS.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstddef>
 #include <tuple>
 #include <vector>
 


src/main/native/include/sysid/analysis/OLS.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstddef>
```
calcmogul, line 7: If you include cstddef, you should be using std::size_t. Doing otherwise is technically undefined behavior (not that the compiler would do anything particularly insane here).

calcmogul:
Heh. wpiformat enforces a condition that's technically UB.

Piphi5:
Should that be a wpiformat fix or should I just go back to using cstddef with size_t?

calcmogul:
Idk. I'm asking on Slack.

calcmogul:
After digging through the C++ standard, it claims cstdint and cstddef should define the types and macros the same as stdint.h and stddef.h respectively. Therefore, we should be OK including cstddef while using size_t.

calcmogul (approved):

PR #140, opened 2021-05-21
Simplify final data preparation

Piphi5:
Reduces verbosity and redundancy of assigning datasets.

### DIFF
--- a/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -87,6 +87,36 @@ static std::vector<PreparedData> Concatenate(
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -86,6 +86,24 @@ static std::vector<PreparedData> Concatenate(
   return dest;
 }
 
+static void StoreDatasets(wpi::StringMap<Storage>* dataset,
```
calcmogul, line 89: Needs a doxygen comment.

calcmogul (changes requested):

calcmogul (approved):

calcmogul:
It worked on a couple JSONs I tried locally as well.

PR #141, opened 2021-05-22
Embed robot program into main application

prateekma:
Heavily refactors the build logic to reduce duplication and enable proper dependency management which is needed to build the fat application.


Piphi5:
Will these changes allow us to deploy from the application itself?

prateekma:
Probably not in this PR. This PR is mainly to actually embed the executable in the application itself. The deployment of the embedded application will be left for another PR.

calcmogul (approved):

PR #142, opened 2021-05-23
Set a maximum data collection duration

### LINKED ISSUE #55: UI freezes if robot program sends a lot of data
Left my Romi spinning for 2 minutes on the quasistatic forward test, and it hard froze my desktop for 30s

Piphi5:
To prevent issues such as #55 and https://github.com/wpilibsuite/frc-characterization/issues/195, the vector storing the collected data now stops after a certain threshold is reached.

Fixes #55 

The sample drivetrain test below was run for several minutes, but only the first 20 seconds of the data were sent.
image

### DIFF
--- a/sysid-application/src/integrationtest_analysis/native/cpp/AnalysisTest.cpp
+++ b/sysid-application/src/integrationtest_analysis/native/cpp/AnalysisTest.cpp
@@ -30,6 +30,9 @@ constexpr double Kv = 1.98;
 constexpr double Ka = 0.2;
 constexpr double kTrackWidth = 0.762;
 
+// 36000 max samples / 9 samples per entry
+constexpr size_t kMaxDataSize = 4000;
+
 // Calculated by finding the voltage required to hold the arm horizontal in the
 // simulation program.
 constexpr double kCos = .250;
@@ -46,6 +49,7 @@ class AnalysisTest : public ::testing::Test {
     m_nt = nt::GetDefaultInstance();
     m_enable = nt::GetEntry(m_nt, "/SmartDashboard/SysIdRun");
     m_kill = nt::GetEntry(m_nt, "/SmartDashboard/SysIdKill");
+    m_overflow = nt::GetEntry(m_nt, "/SmartDashboard/SysIdOverflow");
 
     // Setup logger.
     m_logger.SetLogger([](unsigned int level, const char* file,
@@ -73,7 +77,7 @@ class AnalysisTest : public ::testing::Test {
     }
   }
 
-  void TearDown() override {
+  void AnalyzeJSON() {
     // Save the JSON and make sure that everything checks out.
     auto path = m_manager->SaveJSON(EXPAND_STRINGIZE(PROJECT_ROOT_DIR));
     try {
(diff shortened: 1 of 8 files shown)


sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
```diff
@@ -129,6 +133,12 @@ void TelemetryManager::EndTest() {
                    m_settings.unitsPerRotation;
         stream << "The encoder reported traveling " << p << " " << units << ".";
       }
+
+      if (m_params.overflow) {
+        stream << "NOTE: the robot stopped recording data early because the "
```
calcmogul, line 138: Looks like there should be a linebreak before this note. Also, the contents should be placed in a wrapped textbox so it doesn't go past the end of the box as in the screenshot.

Piphi5:
Here's the new popup
image

sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
```diff
@@ -81,6 +83,8 @@ void TelemetryManager::BeginTest(wpi::StringRef name) {
                     nt::Value::MakeString(m_settings.mechanism.name));
   // Clear the telemetry entry
   nt::SetEntryValue(m_telemetry, nt::Value::MakeString(""));
+  // Set Overflow to False
+  nt::SetEntryValue(m_telemetry, nt::Value::MakeBoolean(false));
```
calcmogul, line 87:
```suggestion
  nt::SetEntryValue(m_overflow, nt::Value::MakeBoolean(false));
```

calcmogul (changes requested):

Piphi5:
Would it be worth adding additional asserts to make sure that the overflow flag is properly set and an additional Analysis test to make sure the flag is thrown?

sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
```diff
@@ -129,6 +133,12 @@ void TelemetryManager::EndTest() {
                    m_settings.unitsPerRotation;
         stream << "The encoder reported traveling " << p << " " << units << ".";
       }
+
+      if (m_params.overflow) {
+        stream << "\nNOTE: the robot stopped recording data early because the "
+                  "maximum amount of entries that could be "
+                  "stored were reached.";
```
calcmogul, line 140:
```suggestion
        stream << "\nNOTE: the robot stopped recording data early because the "
                  "entry storage was exceeded.";
```

calcmogul:
Yea, I think adding those checks would be worth it.

sysid-application/src/integrationtest_analysis/native/cpp/AnalysisTest.cpp
```diff
@@ -146,42 +150,64 @@ class AnalysisTest : public ::testing::Test {
     std::system(del.c_str());
   }
 
+  void OverflowTest() {
```
calcmogul, line 153: The `Test` suffix in `OverflowTest` implies it's a test case, but it's really just a subroutine.
```suggestion
  void VerifyOverflow() {
```

calcmogul (approved):

Issue #114, opened 2021-04-26
Add Romi Vendordep to Project Generation


Piphi5:
Currently, the Gyro part of the Romi project is waiting on the Romi vendordep. Once that's out, it needs to be added to the code.

PR #144, opened 2021-05-27
General C++ cleanups

calcmogul:
* Replace wpi::StringRef with std::string_view
* Replace std::stringstream with fmt::format()
* Remove unused includes
* Replace wpi::math::pi with wpi::numbers::pi

### DIFF
--- a/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
+++ b/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
@@ -35,7 +35,8 @@ wpi::StringMap<wpi::SmallVector<std::string, 4>> gyroCtorMap = {
     {"AnalogGyro", kAnalogCtors},
     {"Pigeon", kPigeonCtors},
     {"ADXRS450", kADXRS450Ctors},
-    {"NavX", kNavXCtors},
+    // FIXME: Waiting on Linux and macOS builds for navX AHRS
+    // {"NavX", kNavXCtors},
     {"None", kAnalogCtors}};
 
 class GenerationTest : public ::testing::Test {
@@ -148,8 +149,9 @@ TEST_F(GenerationTest, Drivetrain) {
     }
   }
 
-  m_settings = sysid::kRomiConfig;
-  m_manager.SaveJSON(m_jsonPath, 1, true);
-  fmt::print(stderr, "Testing: Romi Config\n");
-  Run();
+  // FIXME: Uncomment once Romi vendordep is available
+  // m_settings = sysid::kRomiConfig;
+  // m_manager.SaveJSON(m_jsonPath, 1, true);
+  // fmt::print(stderr, "Testing: Romi Config\n");
+  // Run();
 }
--- a/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
+++ b/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
@@ -4,20 +4,16 @@
 
 #include <cstdlib>
 #include <exception>
-#include <fstream>
-#include <iostream>
-#include <map>
+#include <string>
+#include <string_view>
 #include <thread>
 
+#include <fmt/core.h>
 #include <ntcore_c.h>
 #include <ntcore_cpp.h>
-#include <wpi/FileSystem.h>
 #include <wpi/Logger.h>
-#include <wpi/SmallString.h>
 #include <wpi/SmallVector.h>
 #include <wpi/StringMap.h>
-#include <wpi/StringRef.h>
-#include <wpi/raw_ostream.h>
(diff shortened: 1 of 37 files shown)


sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
```diff
@@ -145,13 +135,13 @@ TEST_F(GenerationTest, Drivetrain) {
   // Encoders + Motor controllers already tested, now just need to test the
   // gyros
   for (auto&& gyro : sysid::kGyros) {
-    if (wpi::StringRef(gyro) == "Romi") {
+    if (std::string_view{gyro} == "Romi") {
```
Piphi5, line 138: We could probably also skip the NavX until we actually get the binaries for it since it seems to crash on the [NavX](https://github.com/wpilibsuite/sysid/pull/144/checks?check_run_id=2702756388).

sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
```diff
@@ -145,13 +135,13 @@ TEST_F(GenerationTest, Drivetrain) {
...
       continue;
     }
     m_settings.m_gyro = gyro;
     for (auto&& gyroCtor : gyroCtorMap[gyro]) {
       m_settings.m_gyroCtor = gyroCtor;
-      wpi::outs() << "Testing: " << gyro << " using " << gyroCtor << "\n";
+      fmt::print("Testing: {} using {}\n", gyro, gyroCtor);
```
Piphi5, line 144: It also seems like this print isn't being printed out in order: https://github.com/wpilibsuite/sysid/pull/144/checks?check_run_id=2702756388. These should precede the output from deploying the simulation, but it only gets printed at the end.

Piphi5 (changes requested):

Piphi5:
Seems like fmt print isn't printing things in order for some reason: https://github.com/wpilibsuite/sysid/pull/144/checks?check_run_id=2702756367, https://github.com/wpilibsuite/sysid/pull/144/checks?check_run_id=2702756388.

prateekma:
I think you need to explicitly flush in GoogleTest.

calcmogul:
Since they're debug prints anyway, I'll just make them go to stderr so no flushing will be needed.

sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
calcmogul: The original GenerationTest.cpp doesn't crash though.

Piphi5: That's true, but it does seem like the NavX test case is bad.

sysid-projects/drive/src/main/cpp/Robot.cpp
```diff
@@ -169,10 +158,7 @@ Robot::Robot() : frc::TimedRobot(5_ms) {
         // } else if (gyroType == "Romi") {
         //   m_gyro = std::make_unique<frc::RomiGyro>();
       } else {
-        std::stringstream ss(gyroCtor);
-        int port = 0;
-        ss >> port;
-        m_gyro = std::make_unique<frc::AnalogGyro>(port);
+        m_gyro = std::make_unique<frc::AnalogGyro>(std::stoi(gyroCtor));
```
Piphi5, line 161: Testing the NavX would lead to this code running. Maybe stoi crashes when the constructor is not an int whereas stringstream doesn't crash?

calcmogul: Yep. I just added a try-catch that defaults the port number to zero. stderr prints in projects don't show in the console during builds, so there's not much we can do to report errors besides crashing.

calcmogul:
I tried removing navX from the list of generation combos, but std::stoi() still threw exceptions, so it seems that's not the only faulty input.

Piphi5:
Seems like the Romi config would also go to that branch, which passes a constructor of `""`

sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
```diff
@@ -163,6 +150,6 @@ TEST_F(GenerationTest, Drivetrain) {
 
   m_settings = sysid::kRomiConfig;
   m_manager.SaveJSON(m_jsonPath, 1, true);
-  wpi::outs() << "Testing: Romi Config\n";
+  fmt::print(stderr, "Testing: Romi Config\n");
   Run();
```
Piphi5, line 154: Commenting this out would stop the Romi from being tested, then we can add it back once #114 is done

prateekma (approved):

PR #147, opened 2021-05-27
Disable analysis when an exception is thrown

### LINKED ISSUE #74: DARE exception prevents loading JSON and modifying settings to fix it

Piphi5:
Whenever an exception is caught, the tool disables data analysis until a feedforward value is changed. This lets users change the data without having to reload the file.

Fixes #74

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -29,6 +29,8 @@
 
 using namespace sysid;
 
+static double kBadDataGain = -9999.0;
+
 Analyzer::Analyzer(wpi::Logger& logger) : m_logger(logger) {
   // Fill the StringMap with preset values.
   m_presets["Default"] = presets::kDefault;
@@ -44,8 +46,11 @@ Analyzer::Analyzer(wpi::Logger& logger) : m_logger(logger) {
   m_location = glass::GetStorage().GetStringRef("AnalyzerJSONLocation");
 }
 
-static void DisplayGain(const char* text, double* data) {
+void Analyzer::DisplayGain(const char* text, double* data) {
   ImGui::SetNextItemWidth(ImGui::GetFontSize() * 5);
+  if (!m_enabled) {
+    data = &kBadDataGain;
+  }
   ImGui::InputDouble(text, data, 0.0, 0.0, "%.5G",
                      ImGuiInputTextFlags_ReadOnly);
 }
@@ -56,12 +61,14 @@ static void SetPosition(double beginX, double beginY, int xShift, int yShift) {
 }
 
 void Analyzer::ConfigParamsOnFileSelect() {
-  m_stepTestDuration = m_settings.stepTestDuration.to<float>();
+  if (m_enabled) {
+    m_stepTestDuration = m_settings.stepTestDuration.to<float>();
 
-  // Estimate qp as 1/8 * units-per-rot
-  m_settings.lqr.qp = 0.125 * m_manager->GetFactor();
-  // Estimate qv as 1/4 * max velocity = 1/4 * (12V - kS) / kV
-  m_settings.lqr.qv = 0.25 * (12.0 - m_ff[0]) / m_ff[1];
+    // Estimate qp as 1/8 * units-per-rot
+    m_settings.lqr.qp = 0.125 * m_manager->GetFactor();
+    // Estimate qv as 1/4 * max velocity = 1/4 * (12V - kS) / kV
+    m_settings.lqr.qv = 0.25 * (12.0 - m_ff[0]) / m_ff[1];
+  }
 }
 
 void Analyzer::Display() {
@@ -77,10 +84,17 @@ void Analyzer::Display() {
         m_manager = std::make_unique<AnalysisManager>(*m_location, m_settings,
                                                       m_logger);
         m_type = m_manager->GetAnalysisType();
+
         Calculate();
         PrepareData();
         PrepareGraphs();
         ConfigParamsOnFileSelect();
+
+        // Since the data preparation methods catch errors, the error needs to
+        // be rethrown so the logger can report it.
+        if (!m_enabled) {
+          throw std::runtime_error(m_exception);
+        }
       } catch (const std::exception& e) {
         // If we run into an error here, let's just ignore it and make the user
         // explicitly select their file.
@@ -88,7 +102,6 @@ void Analyzer::Display() {
         WPI_ERROR(
             m_logger,
             "An error occurred when attempting to load the previous JSON.");
-        static_cast<void>(e);
       }
     } else {
       *m_location = "";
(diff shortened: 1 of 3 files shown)


Piphi5:
Should this also be done for the Feedback gains?

calcmogul:
Yes.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -485,31 +498,35 @@ void Analyzer::SelectFile() {
...
-    m_exception =
-        "The provided JSON was invalid! You may need to rerun the logger.\n" +
-        std::string(e.what());
-    ResetManagerState();
-    ImGui::OpenPopup("Exception Caught!");
-  } catch (const std::exception& e) {
-    m_exception = e.what();
-    ResetManagerState();
-    ImGui::OpenPopup("Exception Caught!");
+  if (m_enabled) {
```
calcmogul, line 501: You could reduce nesting here by putting this at the top instead:
```cpp
if (!m_enabled) {
  return;
}
```

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -485,31 +498,35 @@ void Analyzer::SelectFile() {
...
   try {
-    const auto& [ff, fb, trackWidth] = m_manager->Calculate();
-    m_ff = std::get<0>(ff);
-    m_rSquared = std::get<1>(ff);
-    m_Kp = fb.Kp;
-    m_Kd = fb.Kd;
-    m_trackWidth = trackWidth;
-    m_unit = m_manager->GetUnit();
-    m_factor = m_manager->GetFactor();
+    if (m_enabled) {
```
calcmogul, line 520: Make the function exit before the try block if analysis is disabled.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -526,17 +543,19 @@ void Analyzer::AbortDataPrep() {
...
-    m_dataThread = std::thread([&] {
-      m_plot.SetData(m_manager->GetRawData(), m_manager->GetFilteredData(),
-                     m_manager->GetUnit(), m_ff, m_manager->GetStartTimes(),
-                     m_type, m_abortDataPrep);
-    });
-  } catch (const std::exception& e) {
-    m_exception = e.what();
-    ResetManagerState();
-    ImGui::OpenPopup("Exception Caught!");
+  if (m_enabled) {
```
calcmogul, line 546: Make the function exit before the try block if analysis is disabled.

calcmogul (changes requested):

calcmogul:
I can't even get nonzero P and D gains on main, so idk how to test this PR.

calcmogul:
If I apply this patch:
```diff
diff --git a/sysid-application/src/main/native/cpp/analysis/FeedbackAnalysis.cpp b/sysid-application/src/main/native/cpp/analysis/FeedbackAnalysis.cpp
index 2732976..ad74f9e 100644
--- a/sysid-application/src/main/native/cpp/analysis/FeedbackAnalysis.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/FeedbackAnalysis.cpp
@@ -31,6 +31,7 @@ FeedbackGains sysid::CalculatePositionFeedbackGains(
     // Create an LQR with 2 states to control -- position and velocity.
     frc::LinearQuadraticRegulator<2, 1> controller{
         system, {params.qp, params.qv}, {params.r}, preset.period};
+    throw std::runtime_error("TEST");
     // Compensate for any latency from sensor measurements, filtering, etc.
     controller.LatencyCompensate(system, preset.period,
                                  preset.positionMeasurementDelay);
@@ -48,6 +49,7 @@ FeedbackGains sysid::CalculatePositionFeedbackGains(
   // Create an LQR with one state -- position.
   frc::LinearQuadraticRegulator<1, 1> controller{
       system, {params.qp}, {params.r}, preset.period};
+  throw std::runtime_error("TEST");
   // Compensate for any latency from sensor measurements, filtering, etc.
   controller.LatencyCompensate(system, preset.period,
                                preset.positionMeasurementDelay);
@@ -72,6 +74,7 @@ FeedbackGains sysid::CalculateVelocityFeedbackGains(
   // Create an LQR controller with 1 state -- velocity.
   frc::LinearQuadraticRegulator<1, 1> controller{
       system, {params.qv}, {params.r}, preset.period};
+  throw std::runtime_error("TEST");
   // Compensate for any latency from sensor measurements, filtering, etc.
   controller.LatencyCompensate(system, preset.period,
                                preset.velocityMeasurementDelay);
```
With this data: [noisy-ci.json.txt](https://github.com/wpilibsuite/sysid/files/6558888/noisy-ci.json.txt)

sysid crashes instead of catching the exception. `main` does this too.

Piphi5:
I can't seem to reproduce this issue, it catches it for me.

Piphi5:
It segfaults when we try to automatically calculate the qv and qp for the LQR since it doesn't have any valid data to go off of. Fix will be pushed soon.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -29,6 +29,8 @@
 
 using namespace sysid;
 
+static double badDataGain = -9999.0;
```
calcmogul, line 32:
```suggestion
static constexpr double kBadDataGain = -9999.0;
```

calcmogul:
That fixed the crash, but I'm noticing that when LQR throws an exception, the GUI just doesn't show anything and doesn't open a pop-up with the exception body.

Piphi5:
Is that when you apply the previous patch and load the noisy data?

calcmogul:
Yea

sysid-application/src/main/native/cpp/view/Analyzer.cpp
Piphi5: The compiler doesn't like the fact that I try to convert the const double* to a regular double* for the InputDouble method. Is there any way to work around that?

calcmogul: Oh, ok.

Piphi5:
When files are initially read, any popup that is opened would only be opened once, therefore not staying around long enough for someone to see it. So now the popup is triggered by a boolean rather than opening it to work around this.

calcmogul (approved):

Issue #148, opened 2021-05-28
Add Debug Level Logging to help debug segfaults


Piphi5:
Having a debug level log would make it much easier for developers to debug any segfaults/crashes by being able to quickly identify when they occurred.

PR #153, opened 2021-06-02
Define NDEBUG in release mode

prateekma:
For debug messages to be hidden in release mode, wpi::Logger requires
NDEBUG to be defined.


Piphi5:
Will this also print debug messages onto stdout to help with issue #148

prateekma:
Not to stdout, but the program log. Of course, you can configure the logger to print to stdout if you want it to.

PR #155, opened 2021-06-03
Increase ArmSim error tolerance

### LINKED ISSUE #134: Bad Data Takes way too long to load
I tried running the analyzer for 5+ minutes and it didn't finish:
[sysid_data20210424-095550.json.txt](https://github.com/wpilibsuite/sysid/files/6449686/sysid_data20210424-095550.json.txt)

calcmogul:
This makes the arm sim for the data in #134 complete in about 5 seconds on my
machine with a debug build. Release builds are instantaneous.

Fixes #134.


Piphi5:
We don't have to worry about the other tests since they are linear systems right?

calcmogul:
Correct.

prateekma (approved):

PR #156, opened 2021-06-03
Add angular drivetrain tests

### LINKED ISSUE #149: Finalize the Angular Drivetrain Test
The angular drivetrain hasn't been finalized nor integrated into the analysis tests.

Piphi5:
This works with the current drivetrain simulation model in wpilib although it will probably be updated soon.
Fixes #149

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -11,6 +11,7 @@
 #include <functional>
 #include <initializer_list>
 #include <stdexcept>
+#include <string_view>
 #include <vector>
 
 #include <fmt/format.h>
@@ -81,6 +82,22 @@ static std::vector<PreparedData> Concatenate(
 
   // Return the dest vector.
   return dest;
+}
+
+template <size_t S>
+static void CopyRawData(
+    wpi::StringMap<std::vector<std::array<double, S>>>* dataset) {
+  using Data = std::array<double, S>;
+  auto& data = *dataset;
+  // Loads the Raw Data
+  sysid::ApplyToData(
+      data,
+      [&](std::string_view key) {
+        data[fmt::format("raw-{}", key)] = std::vector<Data>(data[key]);
+      },
+      [](std::string_view key) {
+        return key.find("raw") == std::string_view::npos;
+      });
 }
 
 /**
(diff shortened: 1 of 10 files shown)


Piphi5:
Now uses angular gyro data rather than velocities.

Piphi5:
Should there be unit tests for the angular analysis or does the integration test work fine?

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -247,17 +258,24 @@ static void PrepareAngularDrivetrainData(
...
     for (auto&& pt : it->second) {
-      pt[kVoltageCol] = 2 * std::copysign(pt[kVoltageCol], pt[kAngularRateCol]);
       pt[kLPosCol] *= factor;
       pt[kRPosCol] *= factor;
+
+      // Store the summarized voltages and velocities in the left column
+      pt[kLVoltageCol] =
+          (std::copysign(pt[kLVoltageCol], pt[kAngularRateCol]) +
+           std::copysign(pt[kRVoltageCol], pt[kAngularRateCol])) /
+          2;
```
calcmogul, line 268: The comment is wrong. The velocity isn't being stored in the left column.

calcmogul (changes requested):
Could you post a screenshot of the changes to GUI created by DisplayFeedforwardGains() and some combined plots from an angular test?

calcmogul:
There should be unit tests for all types of analysis, both linear and angular.

Piphi5:
Here's the added spacing for track width:
image
Here's a combined plot:
image

sysid-application/src/main/native/include/sysid/analysis/AngularDriveSim.h
```diff
@@ -0,0 +1,66 @@
...
+ * purposes.
+ */
+class AngularDriveSim {
+ public:
+  /**
+   * @param Ks              Static friction gain.
+   * @param Kv              Velocity gain.
+   * @param Ka              Acceleration gain.
+   * @param initialPosition Initial elevator position.
+   * @param initialVelocity Initial elevator velocity.
```
calcmogul, line 23: Is this linear or angular position/velocity? The model states should be documented.

calcmogul (changes requested):

sysid-application/src/main/native/include/sysid/analysis/AngularDriveSim.h
Piphi5: Is the new explanation better?

calcmogul: Do Kv and Ka also use angular units here?

Piphi5: No, they are linear. Currently, the model takes the linear gains converted into angular gains, and then when generating the data, it spits out the linear gains again (effectively canceling out the conversions). Would it just be better to start with the linear gains and just spit those out?

calcmogul: The linear ones are fine. Just document the units.

calcmogul (approved):

PR #157, opened 2021-06-04
Add robot deploy infrastructure

prateekma:
- Added primary deploy infrastructure
 - Refactored ConfigManager and Generator GUI
 - Added convenience method to save files and use it wherever possible.


sysid-application/src/main/native/cpp/view/Generator.cpp
```diff
@@ -287,6 +308,71 @@ void Generator::Display() {
   }
 
   ImGui::Separator();
+
+  // Create the deploy button.
+  if (ImGui::Button("Deploy to Robot")) {
+    // Acquire a deploy session and start it in the event loop runner.
+    m_deployRunner.ExecSync([this](wpi::uv::Loop& lp) {
+      // FIXME: drive? and json parameters.
+      m_deploySession = DeploySession::Acquire(lp, *m_pTeam, true, wpi::json{},
```
prateekma, line 317: Need to split acquire and exec because this does not run on UI thread.

sysid-application/src/main/native/cpp/Util.cpp
```diff
@@ -52,3 +54,21 @@ std::string sysid::GetAbbreviation(std::string_view unit) {
...
+  wpi::raw_fd_ostream ostream{path.string(), ec};
+
+  // Check error code.
+  if (ec) {
+    throw std::runtime_error("Cannot write to file: " + ec.message());
+  }
+
+  // Write contents.
+  ostream << contents;
+  ostream.flush();
```
calcmogul, line 73: The destructor automatically flushes the file, so the call to flush here is redundant.
```cpp
raw_fd_ostream::~raw_fd_ostream() {
  if (FD >= 0) {
    flush();
    if (ShouldClose && ::close(FD) < 0)
      error_detected(std::error_code(errno, std::generic_category()));
  }
```

calcmogul (changes requested):

calcmogul (dismissed):

PR #158, opened 2021-06-07
Use fmtlib and string_view throughout

prateekma:


sysid-application/src/main/native/cpp/analysis/JSONConverter.cpp
```diff
@@ -117,18 +119,17 @@ std::string sysid::ToCSV(std::string_view path, wpi::Logger& logger) {
...
-               << "Left Velocity (" << unit << "/s),"
-               << "Right Velocity (" << unit << "/s),"
-               << "Gyro Position (deg),Gyro Rate (deg/s)";
-
+    fmt::print(
+        outputFile,
+        "Left Volts (V),Right Volts (V),Left Position ({0}),Right "
+        "Position ({0}),Left Velocity ({0}/s),Right Velocity ({0}/s),Gyro "
+        "Position (deg),Gyro Rate (deg/s)\n",
+        abbreviation);
```
PeterJohnson, line 129: Did you mean to use unit rather than abbreviation for some of these?

prateekma: No, they should all be abbreviation; not sure why it was unit before. Abbreviation is used for velocity in the second case as well.

PR #160, opened 2021-06-07
Mark a test as "Not Run" if no data is collected

### LINKED ISSUE #150: Running a test without collecting any data registers as a "runned test" in the logger
Currently, you can start a test and then end it without enabling the robot and the logger doesn't keep the test at "Not Run".

Piphi5:
Fixes #150

### DIFF
--- a/sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
+++ b/sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
@@ -112,12 +112,11 @@ void TelemetryManager::EndTest() {
 
   // Call the cancellation callbacks.
   for (auto&& func : m_callbacks) {
+    std::string msg;
     if (!m_params.data.empty()) {
       std::string units;
       std::transform(m_settings.units.begin(), m_settings.units.end(),
                      units.begin(), ::tolower);
-
-      std::string msg;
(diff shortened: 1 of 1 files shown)


sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
```diff
@@ -141,8 +140,16 @@ void TelemetryManager::EndTest() {
             "\nNOTE: the robot stopped recording data early because the entry "
             "storage was exceeded.";
       }
-      func(msg);
+
+    } else {
+      msg = "No data was detected.";
     }
+    func(msg);
```
calcmogul, line 147: What does func() do?

Piphi5: It's a function that displays any telemetry messages onto the GUI

calcmogul:
I'm not sure how to test this PR.

Piphi5:
I tested it by connecting to the analysis-test project and then starting and canceling the quasistatic test and making sure that the test was set to `Not Run` after it was prematurely canceled.

calcmogul (approved):
I confirmed main has the issue and this PR fixes it.

PR #164, opened 2021-06-11
Add Debug Prints to console

### LINKED ISSUE #148: Add Debug Level Logging to help debug segfaults

Piphi5:
This should help identify segfaults when they occur from new changes, particularly to the analysis part.

Fixes #148

### DIFF
--- a/sysid-application/src/main/native/cpp/App.cpp
+++ b/sysid-application/src/main/native/cpp/App.cpp
@@ -2,6 +2,8 @@
 // Open Source Software; you can modify and/or share it under the terms of
 // the WPILib BSD license file in the root directory of this project.
 
+#include <cstdio>
+
 #ifndef RUNNING_SYSID_TESTS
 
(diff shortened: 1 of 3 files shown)


sysid-application/src/main/native/cpp/App.cpp
```diff
@@ -2,6 +2,8 @@
 // Open Source Software; you can modify and/or share it under the terms of
 // the WPILib BSD license file in the root directory of this project.
 
+#include <cstdio>
```
calcmogul, line 5: Why is cstdio being included?

Piphi5: Not sure, maybe because of the `stderr`?

calcmogul (approved):

PR #165, opened 2021-06-22
Revert "Upgrade to WPILib's Ubuntu 20.04 docker container"

prateekma:
Reverts wpilibsuite/sysid#162


Piphi5:
Why was this needed?

ThadHouse:
To work around a GCC-7 bug, but we found how to get GCC-8 on 18.04 and to have it work with Gradle.

PR #180, opened 2021-09-13
Fix Step Voltage default data trim

### LINKED ISSUE #175: Program crashes when loading data file
Attached the data file. This was saved after doing all 4 tests. 

[sysid_data20210822-104822.zip](https://github.com/wpilibsuite/sysid/files/7027725/sysid_data20210822-104822.zip)

I don't see any kind of log or other information that I can include here, is there a way to generate debugging info?

Piphi5:
Fixes #175

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
@@ -91,15 +91,18 @@ units::second_t sysid::TrimStepVoltageData(std::vector<PreparedData>* data,
         GetAccelNoiseFloor(*data, settings->windowSize);
     // Find latest element with nonzero acceleration
     auto endIt = std::find_if(
-        std::reverse_iterator{data->end()},
-        std::reverse_iterator{data->begin()}, [&](const PreparedData& entry) {
+        data->rbegin(), data->rend(), [&](const PreparedData& entry) {
           return std::abs(entry.acceleration) > accelNoiseFloor;
         });
 
-    // Calculate default duration
-    settings->stepTestDuration =
-        std::min(endIt->timestamp - data->front().timestamp + minStepTime + 1_s,
-                 maxStepTime);
+    if (endIt != data->rend()) {
+      // Calculate default duration
+      settings->stepTestDuration = std::min(
+          endIt->timestamp - data->front().timestamp + minStepTime + 1_s,
+          maxStepTime);
+    } else {
+      settings->stepTestDuration = maxStepTime;
+    }
   }
 
   // Find first entry greater than the step test duration


sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -96,10 +95,14 @@ units::second_t sysid::TrimStepVoltageData(std::vector<PreparedData>* data,
           return std::abs(entry.acceleration) > accelNoiseFloor;
         });
 
-    // Calculate default duration
-    settings->stepTestDuration =
-        std::min(endIt->timestamp - data->front().timestamp + minStepTime + 1_s,
-                 maxStepTime);
+    if (endIt != data->rend()) {
```
calcmogul, line 98: Why `rend()` instead of `end()`?

sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -73,7 +73,6 @@ units::second_t sysid::TrimStepVoltageData(std::vector<PreparedData>* data,
                                            units::second_t minStepTime,
                                            units::second_t maxStepTime) {
   auto firstTimestamp = data->at(0).timestamp;
-
```
calcmogul, line 76: This doesn't need to be removed.

sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
```diff
@@ -259,7 +262,6 @@ void sysid::InitialTrimAndFilter(
     AnalysisManager::Settings& settings, units::second_t& minStepTime,
     units::second_t& maxStepTime, std::string_view unit) {
   auto& preparedData = *data;
-
```
calcmogul, line 262: This doesn't need to be removed.

calcmogul (changes requested):

sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
Piphi5: I think endIt is a reverse iterator. Does it not make a difference if I use `end()`?

calcmogul: Oh, I see. In that case, we should probably change this:
```cpp
    auto endIt = std::find_if(
        std::reverse_iterator{data->end()},
        std::reverse_iterator{data->begin()}, [&](const PreparedData& entry) {
          return std::abs(entry.acceleration) > accelNoiseFloor;
        });
```
to this:
```cpp
    auto endIt = std::find_if(
        data->rend(), data->rbegin(), [&](const PreparedData& entry) {
          return std::abs(entry.acceleration) > accelNoiseFloor;
        });
```

PeterJohnson: Wouldn’t that be rbegin(), rend()?  The rbegin() function returns the reverse iterator starting at the end of the container.

Piphi5: Yeah, it is that. It doesn't compile otherwise.

calcmogul (approved):

Piphi5:
We should just double check to make sure the issue is fully resolved on Will's end before merging.

PR #182, opened 2021-09-17
Clean up Eigen usage

calcmogul:
* Replace Matrix<> with Vector<> where vectors are explicitly intended.
  I found these via `rg "Eigen::Matrix<double, \w+, 1>"`.
* Replace MakeMatrix() and operator<< usage with initializer list
  constructors. I found these via `rg MakeMatrix` and `rg "<<"`
  respectively.


calcmogul:
The builds should pass once Eigen 3.4 is merged into allwpilib.

prateekma (approved):

PR #187, opened 2021-10-07
Clean up data processing pipelines

calcmogul:
* Replace ApplyToData() with merged for loops to reduce indirection
  (helps readability)
* Rename AccelAndTimeFilter() to AccelFilter()
* Remove an unused argument to AccelFilter()
* Inline quasistatic data trim since it's a simple one-liner
* Fix the "empty data" check checking the key size (first pair argument)
  instead of the data size (second pair argument)


calcmogul:
Structured bindings would have caught the "size checK" bug, but StringMap doesn't support them.

PR #203, opened 2021-11-11
Replace string_view shims and Split() with wpiutil StringExtras

calcmogul:


calcmogul:
https://en.cppreference.com/w/cpp/string/basic_string_view/contains

sysid-application/src/main/native/include/sysid/Util.h
```diff
@@ -80,7 +80,7 @@ static constexpr bool starts_with(std::string_view obj,
   return obj.substr(0, x.size()) == x;
```
PeterJohnson, line 80: Both starts_with and contains are in wpi/StringExtras.h

sysid-application/src/main/native/cpp/Util.cpp
```diff
@@ -25,13 +24,15 @@ void sysid::CreateTooltip(const char* text) {
...
-  std::string item;
 
-  while (std::getline(ss, item, c)) {
-    result.push_back(item);
+  size_t idx = str.find(separator);
+  result.push_back(std::string{str.substr(0, idx)});
+  while (idx != std::string_view::npos) {
+    str.remove_prefix(idx + 1);
+    idx = str.find(separator);
+    result.push_back(std::string{str.substr(0, idx)});
```
PeterJohnson, line 35: re: split, did you know about https://github.com/wpilibsuite/allwpilib/blob/main/wpiutil/src/main/native/include/wpi/StringExtras.h#L481-L483 ?  Only difference is it's a SmallVector-taking one, rather than std::vector return.

sysid-application/src/main/native/cpp/telemetry/TelemetryManager.cpp
```diff
@@ -249,13 +251,17 @@ void TelemetryManager::Update() {
...
+      wpi::split(m_params.raw, res, ',');
 
       // Convert each string to double.
       std::vector<double> values;
       values.reserve(res.size());
       for (auto&& str : res) {
-        values.push_back(std::stod(str));
+        double value;
+        auto [ptr,
+              ec]{std::from_chars(str.data(), str.data() + str.size(), value)};
```
PeterJohnson, line 263: Most compilers do not have from_chars floating point support.  See https://en.cppreference.com/w/cpp/compiler_support/17

PeterJohnson: You can use wpiutil's parse_float instead.  https://github.com/wpilibsuite/allwpilib/blob/main/wpiutil/src/main/native/include/wpi/StringExtras.h#L659-L660

PeterJohnson (approved):

PR #205, opened 2021-11-12
Clean up motor selection UI

calcmogul:
* Make CAN motor controllers use "Motor CAN ID" instead of "Motor Port"
* Remove the index from "Motor Port i" because the extra number is
  confusing
* Reorder the fields to match encoder selection
* Don't use ternary operators in fmt::format() calls to future-proof
  against compile-time format strings in C++20


calcmogul:
motor-selection-ui-general-mechanism
motor-selection-ui-drivetrain

prateekma (approved):

Issue #195, opened 2021-10-26
Add tests to make sure Generation configures the right hardware


Piphi5:

calcmogul:
Fixed by #211.

Piphi5:
We still need tests on the generated projects end

PR #207, opened 2021-11-13
Fix general mechanism config loading

calcmogul:
The gyro type is saved as "AnalogGyro", but the arrays used for config
loading expect "Analog".

### DIFF
--- a/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
+++ b/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
@@ -46,18 +46,18 @@ wpi::StringMap<wpi::SmallVector<std::string_view, 2>>
 };
 
 const wpi::SmallVector<std::string_view, 4> kPigeonCtors{"0", "WPI_TalonSRX-1"};
-const wpi::SmallVector<std::string_view, 4> kAnalogCtors{"0"};
+const wpi::SmallVector<std::string_view, 4> kAnalogGyroCtors{"0"};
 const wpi::SmallVector<std::string_view, 4> kNavXCtors{
     "SerialPort.kUSB", "I2C", "SerialPort.kMXP", "SPI.kMXP"};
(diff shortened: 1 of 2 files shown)


calcmogul:
Configs exported with the wrong value can be loaded by changing the JSON's "gyro" value from "AnalogGyro" to "Analog".

Piphi5:
Wouldn't AnalogGyro be clearer and so we should update the Robot code?

calcmogul:
The class names don't line up for others either:

|GUI|Constructor name|
|---|----------------|
|Analog|AnalogGyro|
|NavX|AHRS|

calcmogul:
We should consider refactoring this to not use strings so it's less error-prone, but I don't know what that would look like.

Piphi5:
Personally, I think AnalogGyro would be clearer in the GUI instead of passing it as Analog.

Piphi5:
> We should consider refactoring this to not use strings so it's less error-prone, but I don't know what that would look like.
If we setup the #195 tests we wouldn't have to worry about it as much.

calcmogul:
> Maybe to not use strings we could also save the string in the json (for clarity) but also send a number that is mapped out to the right gyro in some file that's compiled with both the application and the project.

That sounds more complicated than what we're doing now.

Piphi5:
Yeah it is. I think resolving #195 should fix things.

Piphi5:
@calcmogul shouldn't the PR be titled "Fix Drivetrain config loading" or "Fix config loading"

Piphi5:
Also, I just realized I misinterpreted the issue. We could create enums with String representations then have methods to convert between them similar to how the [analyzer](https://github.com/wpilibsuite/sysid/blob/main/sysid-application/src/main/native/include/sysid/analysis/AnalysisType.h) differentiates between different mechanism types.

Piphi5 (changes requested):
Keep "AnalogGyro", update the other arrays to store that instead.

Piphi5 (approved):

calcmogul:
Superseded by #211.

PR #216, opened 2021-11-19
Set default gyro and encoder idxs to zero

Piphi5:
The default array is of length two leading to weird issues that often
crashed the release build of the program.


calcmogul:
If the default array is 2 elements long and indices are 0-based, 2 is beyond the end of the array.

calcmogul (approved):

PR #220, opened 2021-11-22
Display raw data and let users adjust variables if filtering fails

### LINKED ISSUE #215: Motion threshold is too high and isn't customizable upon JSON load failure
From https://usfirst.collab.net/sf/go/post7071:

> After running the quasistatic and dynamic tests in sysID and then running the analyzer, I get the error message " Quasistatic test trimming removed all data." It is somewhat difficult to diagnose the problem without access to the project generated by sysID and deployed to the robot. Reported encoder values looked fine and the robot was driving forward and backwards correctly. I have attached the data collected from the tests here:
[sysid_data20211116-202237.json.txt](https://github.com/wpilibsuite/sysid/files/7558287/sysid_data20211116-202237.json.txt)

After plotting the data with https://github.com/wpilibsuite/sysid/blob/main/scripts/time_plots.py, it appears the data is below the default motion threshold of 0.2 (the data is around 0.04). We should probably add a more integrated way to view the raw time domain data.

Also, the JSON loading is aborted, which means the motion threshold can't be fixed by the user if there's a failure. We should fix that.

Piphi5:
Fixes #215 

Also removes and filtering of the raw data.

### DIFF
--- a/sysid-application/src/integrationtest-analysis/native/cpp/AnalysisTest.cpp
+++ b/sysid-application/src/integrationtest-analysis/native/cpp/AnalysisTest.cpp
@@ -89,6 +89,8 @@ class AnalysisTest : public ::testing::Test {
       }
       sysid::AnalysisManager analyzer{path, analyzerSettings, m_logger};
 
+      analyzer.PrepareData();
+
       const auto& [ff, fb, trackWidth] = analyzer.Calculate();
       const auto& ffGains = std::get<0>(ff);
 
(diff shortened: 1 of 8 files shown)


calcmogul:
I've confirmed this works with good JSONs, and it allows manually fixing the motion threshold of the bad JSON provided in #215 to get plots. However, if I load the bad JSON first, then a good JSON, then the bad JSON again, it leaves the fit lines from the good JSON instead of clearing them.

[bad-data.json.txt](https://github.com/wpilibsuite/sysid/files/7610714/bad-data.json.txt)
[good-data.json.txt](https://github.com/wpilibsuite/sysid/files/7610715/good-data.json.txt)

Piphi5:
This update fixed that issue for me

calcmogul (approved):

PR #224, opened 2021-11-26
Fix windows generation test issue

### LINKED ISSUE #223: Generation Integration Tests take very long to exit on Windows
As seen in recent Integration Test Runs (e.g. [this](https://github.com/wpilibsuite/sysid/actions/runs/1495725156), [this](https://github.com/wpilibsuite/sysid/actions/runs/1495365408), and [this](https://github.com/wpilibsuite/sysid/actions/runs/1495364473)), it takes about 3 hours to exit the test and it's usually held up by the generation test on Windows (the test finishes but gradle doesn't exit).

I have reproduced the issue on my Windows machine and isolated it to the addition of the gtest stdout capture, however, I haven't found any fixes for it yet.

Piphi5:
Fixes #223 by starting console output capture right before the simulation window is launched and ending it before it's closed which seems to fix the issue of the process hanging afterwards.

### DIFF
--- a/sysid-application/src/integrationtest-utils/native/cpp/IntegrationUtils.cpp
+++ b/sysid-application/src/integrationtest-utils/native/cpp/IntegrationUtils.cpp
@@ -4,6 +4,7 @@
 
 #include "IntegrationUtils.h"
 
+#include <stdexcept>
 #include <string>
 
 #include <fmt/core.h>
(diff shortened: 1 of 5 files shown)


sysid-application/src/integrationtest-utils/native/cpp/IntegrationUtils.cpp
```diff
@@ -58,7 +62,11 @@ void Connect(NT_Inst nt, NT_Entry kill) {
   }
 }
 
-void KillNT(NT_Inst nt, NT_Entry kill) {
+void KillNT(NT_Inst nt, NT_Entry kill, std::string* logContent) {
+  // Before killing sim, store any captured console output.
+  auto& capturedOutput = *logContent;
+  capturedOutput = ::testing::internal::GetCapturedStdout();
```
calcmogul, line 68: You could store the std::string, then return it by value instead. That means the caller gets their own copy, RVO kicks in, and you won't need the static global string.

calcmogul (changes requested):

calcmogul (approved):

PR #226, opened 2021-11-27
Use new Glass Storage API

calcmogul:


.gitignore
```diff
@@ -300,3 +300,7 @@ compile_commands.json
 
 # Generation Config
 config.json
+
+# ImGUI JSONs
+imgui-window.json
+imgui.json
```
PeterJohnson: This can be changed to sysid.json / sysid-window.json

calcmogul:
Should they even be in .gitignore since the files go in ~/.config now?

PeterJohnson:
Probably not.  Unless for testing purposes you're running `sysid .` or something.

PeterJohnson (approved):

PR #228, opened 2021-11-28
Reorder variable naming for datasets

Piphi5:


calcmogul:
You should update the Doxygen for those functions while you're at it.

calcmogul (approved):

PR #229, opened 2021-11-28
Split Gearing into numerator and denominator

Piphi5:


Piphi5 (commented):
Needs to do the same for the Analyzer gearing

Piphi5:
Sorry wrong branch

sysid-application/src/main/native/cpp/view/Generator.cpp
```diff
@@ -516,7 +516,21 @@ void Generator::Display() {
...
+  ImGui::SameLine();
+  ImGui::SetNextItemWidth(ImGui::GetFontSize() * 4);
+  ImGui::InputDouble("##2", &m_settings.gearingDenominator, 0.0, 0.0, "%.2f");
+  ImGui::SameLine();
+  ImGui::Text("Gearing");
+
+  // Ensure no division by zero
+  if (m_settings.gearingDenominator <= 0) {
+    m_gearingDenominator = 1;
+  }
```
calcmogul, line 532: We should do this check for the numerator as well since a zero or negative value won't do sane things either.

calcmogul (changes requested):

sysid-application/src/main/native/cpp/view/Generator.cpp
```diff
@@ -516,7 +516,25 @@ void Generator::Display() {
...
+  ImGui::Text("Gearing");
+
+  // Ensure no division by zero or weird gearing behaviour
+  if (m_settings.gearingDenominator <= 0) {
+    m_settings.gearingDenominator = 1.0;
+  }
+
+  if (m_settings.gearingNumerator <= 0) {
+    m_settings.gearingNumerator = 1.0;
+  }
```
calcmogul, line 536: Flip the order of these since everywhere else uses numerator then denominator.

calcmogul (approved):

PR #231, opened 2021-11-29
Complete doxygen documentation and add gradle task

Piphi5:


sysid-application/src/main/native/cpp/analysis/AngularDriveSim.cpp
```diff
@@ -10,9 +10,10 @@
 
 using namespace sysid;
 
-AngularDriveSim::AngularDriveSim(double kS, double kV, double kA,
-                                 double trackwidth, double initialPosition,
-                                 double initialVelocity)
+AngularDriveSim::AngularDriveSim(double Ks, double Kv, double Ka,
+                                 double trackwidth,
+                                 double initialPosition = 0.0,
+                                 double initialVelocity = 0.0)
```
calcmogul, line 16: Default arguments go in the declaration, not the definition.

sysid-application/src/main/native/include/sysid/analysis/FeedbackControllerPreset.h
```diff
@@ -31,15 +31,23 @@ struct FeedbackControllerPreset {
   /** The measurement delay in the encoder measurements. */
   units::second_t measurementDelay;
 
-  /** Checks equality between two feedback controller presets. */
+  /** Checks equality between two feedback controller presets.
+   * @param rhs Another FeedbackControllerPreset
+   *
+   * @return If the two presets are equal
+   */
```
calcmogul, line 38: This applies to other comments too.
```suggestion
  /**
   * Checks equality between two feedback controller presets.
   *
   * @param rhs Another FeedbackControllerPreset
   * @return If the two presets are equal
   */
```

calcmogul (changes requested):
Needs a rebase as well.

calcmogul (approved):

PR #236, opened 2021-12-04
Upgrade to ImGui 1.85

Piphi5:


calcmogul:
For context, GCC 9 gave an ICE on Eigen stuff.

.github/workflows/integration-test.yml
```diff
@@ -42,6 +42,15 @@ jobs:
...
           architecture: ${{ matrix.architecture }}
 
+      - name: Upgrade GCC
+        run: |
+          sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+          sudo apt install -y gcc-11 g++-11
+          sudo update-alternatives \
+            --install /usr/bin/gcc gcc /usr/bin/gcc-11 11 \
+            --slave /usr/bin/g++ g++ /usr/bin/g++-11
+          sudo update-alternatives --set gcc /usr/bin/gcc-11
```
calcmogul, line 52:
```suggestion
          sudo update-alternatives --set gcc /usr/bin/gcc-11
        if: startsWith(matrix.artifact-name, 'Linux')
```

sciencewhiz:
Does upgrading gcc limit running sysID to very modern distros?

Piphi5:
> Does upgrading gcc limit running sysID to very modern distros?

I'm not entirely sure. However, there's a compiler bug with GCC 9 that doesn't let Eigen compile.

PeterJohnson:
Please break into two PRs (one for imgui and one for GCC).  Changing GCC will make it require a newer libstdc++ be installed.  I would recommend we use the lowest version of GCC that works (in this case, it sounds like that would be 10?).

Piphi5:
The GCC update has now been moved to #237

Piphi5:
It seems like this still builds. Do we know the GCC compiler for the Github Actions machines?

sciencewhiz:
Since we use the wpilib Ubuntu 18.04 docker image for linux, it should be GCC 8. https://github.com/wpilibsuite/docker-images/blob/26ff4a073ef08859ebfd0e9d6ce35236791813d9/ubuntu-base/Dockerfile.18.04#L18

Piphi5:
I just confirmed that it doesn't run into the bug with GCC 8. Maybe GCC 10 is already installed on the actions machines and so that's why it said GCC 10 was installed [here](https://github.com/wpilibsuite/sysid/runs/4417994843?check_suite_focus=true), but it compiled with GCC 8

calcmogul:
I got the following abort after startup with the linux x86-64 artifact.
```
[tav@myriad x86-64]$ ./sysid 
sysid: /__w/thirdparty-imgui/thirdparty-imgui/implot/implot.cpp:2050: void ImPlot::SetupAxisLimits(ImAxis, double, double, ImPlotCond): Assertion `(GImPlot->CurrentPlot != __null && !GImPlot->CurrentPlot->SetupLocked) && "Setup needs to be called after BeginPlot and before any setup locking functions (e.g. PlotX)!"' failed.
Aborted (core dumped)
```

calcmogul (approved):

sciencewhiz:
The integration tests look like they run in a different environment then the builds. The builds run on the wpilib docker container, the integration tests look like they run straight on actions, so they would use the ubuntu 20.04 defaults.

PR #237, opened 2021-12-04
Build Linux with GCC 10

Piphi5:
This helps fix a compiler bug with Eigen


Piphi5:
Looks like GCC 10 is already installed on Github actions

Piphi5:
Do note that the issue is only present with GCC 9. GCC 8, 10, 11 all compile fine.

PR #241, opened 2021-12-05
Add CTRE + ADXRS450 Support and finalize Gyro Integration Tests

### LINKED ISSUE #239: Make integration tests not run for vendordep configurations
Few vendors have bothered to support simulation on all platforms, and others don't plan to going forward (navX). For this tool to be any use, we need support for vendors.

If we remove problematic vendors from our integration tests, we should be able to use athena-only vendordeps irl. This reduction in test coverage isn't ideal, but we've exhausted our options. At this point, it's the vendor's fault if they didn't provide us sim builds and sysid breaks for their vendordep irl.

Piphi5:
Fixes #239

### DIFF
--- a/sysid-library/src/main/include/sysid/generation/SysIdSetup.h
+++ b/sysid-library/src/main/include/sysid/generation/SysIdSetup.h
@@ -4,14 +4,18 @@
 
 #pragma once
 
-// #include <ctre/Phoenix.h>
+#ifdef __FRC_ROBORIO__
+#include <ctre/Phoenix.h>
+#endif
 
 #include <functional>
 #include <memory>
+#include <string>
 #include <string_view>
 #include <vector>
 
 #include <frc/Encoder.h>
+#include <frc/interfaces/Gyro.h>
 #include <frc/motorcontrol/MotorController.h>
 #include <rev/CANSparkMax.h>
 #include <units/voltage.h>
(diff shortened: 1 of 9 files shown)


sysid-library/src/main/include/sysid/generation/SysIdSetup.h
```diff
@@ -86,10 +90,63 @@ void SetupEncoders(
...
     std::unique_ptr<rev::SparkMaxAlternateEncoder>& revDataPort,
     std::unique_ptr<frc::Encoder>& encoder, std::function<double()>& position,
     std::function<double()>& rate);
 
+/**
+ * Sets up an encoder for data collection by settings a position and rate
+ * function to report the right encoder values.
+ *
+ * @param[in] gyroType The type of gyro to configure. Options include: "Analog
+ * Gyro", "ADXRS450", "NavX", "Pigeon", "Romi", "None"
```
calcmogul, line 106:
```suggestion
 * @param[in] gyroType The type of gyro to configure. Options include:
 *                     "Analog Gyro", "ADXRS450", "NavX", "Pigeon", "Romi", "None"
```

sysid-library/src/main/include/sysid/generation/SysIdSetup.h
```diff
@@ -86,10 +90,63 @@ void SetupEncoders(
...
+ * @param[in] gyroCtor The type of constructor to use with the gyro. This varies
+ * based on the gyro.
```
calcmogul, line 108:
```suggestion
 * @param[in] gyroCtor The type of constructor to use with the gyro. This varies
 *                     based on the gyro.
```

calcmogul (changes requested):

calcmogul (approved):

PR #245, opened 2021-12-11
Revert CTRE Addition (Temporary)

Piphi5:


calcmogul:
The gearing fixes should be a separate PR.

Piphi5:
This issue should be resolved

PR #247, opened 2021-12-13
Incorporate gearing into the Spark Max Encoders

Piphi5:


calcmogul:
This change can now be rebased onto main with the CTRE revert commit removed. We merged a fix for CTRE library deploy.

calcmogul (approved):

PR #250, opened 2021-12-15
Fix sysid-projects not using newest allwpilib dev builds

calcmogul:


calcmogul:
It was only pulling 2022.1.1-beta-3 when it should have pulled the commits to main since then as well.

sciencewhiz (commented):
This will work on CI, but will still pull in the latest release installed locally unless you also add `wpi.maven.useLocal = false`

calcmogul:
Do we want it to see local publishes too? idk

sciencewhiz:
That's also needed for local publishes. https://github.com/wpilibsuite/GradleRIO/issues/528#issuecomment-941947942

prateekma (approved):

PR #252, opened 2021-12-16
Build roborio binaries using shared dependencies

ThadHouse:
Vendor dependencies will require shared wpilib. This changes the built binaries to use shared dependencies. Additionally, it generates all the resources for the binaries, and also adds a generated helper function to grab all library binaries. 

It does not currently deploy the binaries, that task is left as an exercise to whoever merges this.


Piphi5:
So someone would have to redo the deploying infrastructure?

ThadHouse:
It won't be a whole rewrite. Just have to add the library deploy.

ThadHouse:
I ended up adding the deploy. And fixed a bug in file deploy with incorrectly using a string_view

prateekma (approved):

PR #256, opened 2021-12-19
Remove CTRE ifdef code for integration tests

Piphi5:
Also fixes a bug caught by the integration tests

### DIFF
--- a/sysid-projects/mechanism/src/main/cpp/Robot.cpp
+++ b/sysid-projects/mechanism/src/main/cpp/Robot.cpp
@@ -50,14 +50,11 @@ Robot::Robot() : frc::TimedRobot(5_ms) {
     }
 
     fmt::print("Initializing encoder\n");
-    sysid::SetupEncoders(
-        encoderType, isEncoding, period, cpr, gearing, numSamples,
-        controllerNames[0], m_controllers.front().get(), encoderInverted,
-        encoderPorts,
-#ifdef __FRC_ROBORIO__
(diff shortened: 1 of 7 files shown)


Piphi5:
Waiting on #255 to be merged

calcmogul:
The commit message should also mention there's changes to some function arguments to fix lifetime issues. Ideally, that would be in a separate commit.

calcmogul:
You should do an interactive rebase with each commit set to "edit" so you can check each commit compiles. That should help you track down the build issue.

sysid-projects/mechanism/src/main/cpp/Robot.cpp
```diff
@@ -50,24 +50,19 @@ Robot::Robot() : frc::TimedRobot(5_ms) {
...
+                         numSamples, controllerNames[0],
+                         m_controllers.front().get(), encoderInverted,
+                         encoderPorts, m_cancoder, m_revEncoderPort,
+                         m_revDataPort, m_encoder, m_position, m_rate);
   } catch (std::exception& e) {
     fmt::print("Project failed: {}\n", e.what());
     std::exit(-1);
   }
   std::cout.flush();
-#ifdef INTEGRATION
```
calcmogul, line 66: This #ifdef shouldn't be removed.

calcmogul (changes requested):

calcmogul (approved):

PR #267, opened 2021-12-24
Shrink project parameter input boxes to fit Units Per Rotation tooltip

### LINKED ISSUE #248: Help Dialog for Units Per Rotation
I am documenting this better but it would help to have a little help be hoverable [here](https://github.com/wpilibsuite/frc-docs/issues/1542).  I think you would have to shrink all of the box sizes slightly but I can't think of anything that would come even close to using that large of a box.
image

Piphi5:
Fixes #248


Piphi5:
New setup:
image

calcmogul (approved):

PR #268, opened 2021-12-24
Display time domain plots above voltage domain plots

### LINKED ISSUE #257: Put time domain plots before voltage domain plots
The time domain plots are more useful for determining goodness of fit in real scenarios than the voltage domain plots.

Piphi5:
Fixes #257

### DIFF
--- a/sysid-application/src/main/native/cpp/view/AnalyzerPlot.cpp
+++ b/sysid-application/src/main/native/cpp/view/AnalyzerPlot.cpp
@@ -511,7 +511,7 @@ bool AnalyzerPlot::DisplayTimeDomainPlots(ImVec2 plotSize) {
 
 bool AnalyzerPlot::LoadPlots() {
   // See if the plots are loaded
-  return DisplayVoltageDomainPlots() && DisplayTimeDomainPlots();
+  return DisplayTimeDomainPlots() && DisplayVoltageDomainPlots();
 }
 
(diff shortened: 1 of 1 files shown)


Piphi5:
Currently, time steps vs time comes before voltage domain. Is this okay or should that be moved after the voltage domain fit?
image

calcmogul:
I think that's fine.

calcmogul (approved):

calcmogul (changes requested):
The combined diagnostics dialog layout got broken.

calcmogul:
Screenshot_20211224_154806

calcmogul:
I think the chart title order is hardcoded and relied on in a bunch of places in AnalyzerPlot.cpp.

Piphi5:
Ah that makes sense

Piphi5:
What if we didn't mess with the order for the combined diagnostics?

calcmogul:
That's not an option without refactoring. We call the same plotting functions for both the normal view and the combined view.

Piphi5:
Well I did also update the order for the combined diagnostics graph, here's what it looks like if I just revert that change:
image

calcmogul:
I guess that's the easiest stop-gap solution. We should refactor that display code at some point though.

Piphi5:
Yeah I'll make a new issue since the refactor is out of the scope of this issue/pr

calcmogul (approved):

PR #279, opened 2021-12-28
Refactor error popups for config JSON loading

### LINKED ISSUE #275: Crash when loading old config
Since this is all in development there probably isn't a guarantee that an old config will work with newer builds. _However_ it still shouldn't crash the program.

Running on Windows 10 64-bit using v2022-1.1-beta4 (btw, you should make it so I can select the text in the info box so I can paste it here).

Loading the attached config crashes the program. I don't see any logs either in AppData or on the console
[mk2_swerve_sterring_config.zip](https://github.com/wpilibsuite/sysid/files/7785719/mk2_swerve_sterring_config.zip)
.

Piphi5:
Fixes #275

### DIFF
--- a/sysid-application/src/main/native/include/sysid/Util.h
+++ b/sysid-application/src/main/native/include/sysid/Util.h
@@ -47,6 +47,14 @@ static constexpr const char* kUnits[] = {"Meters",  "Feet",      "Inches",
 void CreateTooltip(const char* text);
 
 /**
+ * Utility function to launch an error popup if an exception is detected.
+ *
+ * @param isError True if an exception is detected
+ * @param errorMessage The error message associated with the exception
+ */
+void CreateErrorPopup(bool& isError, std::string_view errorMessage);
+
+/**
  * Returns the abbreviation for the unit.
  *
  * @param unit The unit to return the abbreviation for.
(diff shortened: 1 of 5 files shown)


sysid-application/src/main/native/include/sysid/Util.h
```diff
@@ -46,6 +46,8 @@ static constexpr const char* kUnits[] = {"Meters",  "Feet",      "Inches",
  */
 void CreateTooltip(const char* text);
 
+void CreateErrorPopup(bool& isError, std::string_view errorMessage);
+
```
Piphi5, line 50: This should have failed the documentation check

calcmogul: That's because doxygen isn't processing any files. The docs it generates are blank.

calcmogul: https://github.com/wpilibsuite/sysid/pull/283 at least generates something, but it still isn't warning on undocumented functions.

sysid-application/src/main/native/include/sysid/Util.h
```diff
@@ -46,6 +46,8 @@ static constexpr const char* kUnits[] = {"Meters",  "Feet",      "Inches",
  */
 void CreateTooltip(const char* text);
 
+void CreateErrorPopup(bool& isError, std::string_view errorMessage);
```
prateekma, line 49: Can you add docs to this so we can merge?

prateekma (dismissed):

calcmogul (approved):

PR #285, opened 2021-12-29
Improve handling of bad analysis JSONs

### LINKED ISSUE #276: Crash when selecting 'forward' analysis on incomplete dataset, unable to load application after
Attached is just the file after running just the first of the characterization routines. [sysid_data20211228-112415.zip](https://github.com/wpilibsuite/sysid/files/7785854/sysid_data20211228-112415.zip)

When loading this file for analysis, a warning popup comes up saying the data is incomplete. If I then change the analysis dataset from 'combined' to 'forward' the application crashes.

When trying to reopen the application, the following assertion is thrown, and I can no longer open the application.
 
image

The only way to recover is manually modify the "AnalyzerJSONLocation" key in the sysid.json.

Piphi5:
Now if SysId runs into a faulty JSON, it will not store or analyze it. As a result, it fixes #276 and other possible faulty behavior from analyzing incomplete and/or faulty JSON data.

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -95,6 +95,8 @@ void Analyzer::Display() {
         if (!m_enabled) {
           throw std::runtime_error(m_exception);
         }
+      } catch (const wpi::json::exception& e) {
+        HandleJSONError(e);
       } catch (const std::exception& e) {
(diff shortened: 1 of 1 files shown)


calcmogul (approved):
The old version throws an exception when loading the JSON, but still uses the corrupted data. This PR correctly ignores the corrupted JSON data on my Linux machine.

PR #289, opened 2022-01-01
Add Romi to drivetrain project

### LINKED ISSUE #114: Add Romi Vendordep to Project Generation

Piphi5:
This doesn't deploy the simgui so the romi isn't fully supported but its a part of the project and also included in the integration tests.

Fixes #114

### DIFF
--- a/sysid-projects/drive/src/main/cpp/Robot.cpp
+++ b/sysid-projects/drive/src/main/cpp/Robot.cpp
@@ -14,7 +14,6 @@
 #include <fmt/format.h>
 #include <frc/ADXRS450_Gyro.h>
 #include <frc/AnalogGyro.h>
-// #include <frc/romi/RomiGyro.h>
 #include <frc/simulation/DriverStationSim.h>
 #include <frc/smartdashboard/SmartDashboard.h>
 #include <wpi/StringExtras.h>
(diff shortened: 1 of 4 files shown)


sysid-projects/drive/src/main/cpp/Robot.cpp
```diff
@@ -14,7 +14,6 @@
 #include <fmt/format.h>
 #include <frc/ADXRS450_Gyro.h>
 #include <frc/AnalogGyro.h>
-// #include <frc/romi/RomiGyro.h>
```
calcmogul, line 17: Why is this commented out?

Piphi5: We moved all the gyro setup code into `sysid-library` so it wasn't needed in the drivetrain project anymore.

calcmogul: Should it be removed then?

Piphi5: Isn't the git diff saying that it was removed?

calcmogul (approved):

PR #297, opened 2022-01-05
Add gearing to CTRE data collection

Piphi5:
RIght now, CTRE Encoders don't account for Gearing. This fixes it.


calcmogul:
Do any other encoders have this issue?

Piphi5:
They shouldn't. Atleast in the code the roboRIO encoders we're adjusted and the original change was to support the spark max API

calcmogul (approved):

PR #300, opened 2022-01-05
Autoconvert frc-characterization JSONs to SysId format

### LINKED ISSUE #295: Autoconvert frc-characterization JSONs
We currently have a JSON conversion button to convert frc-characterization JSONs. The existing dialog for that is somewhat nonobvious to find and clunky.

The tool knows when it's given an frc-char JSON and the error dialog says to use the converter, so we should just run the converter automatically and display the data as usual.

calcmogul:
Closes #295.

### DIFF
--- a/sysid-application/src/main/native/cpp/App.cpp
+++ b/sysid-application/src/main/native/cpp/App.cpp
@@ -157,12 +157,8 @@ void Application(std::string_view saveDir) {
       ImGui::EndMenu();
     }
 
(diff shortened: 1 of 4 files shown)


calcmogul:
This is equivalent to manually taking any JSON that fails the SysId check, using the JSON conversion button, then feeding the new JSON back in.

PeterJohnson (approved):

prateekma (approved):

PR #304, opened 2022-01-10
Clarify units per rotation tooltip

### LINKED ISSUE #288: Reuse the gearing and analysis type in the Logger
Analysis Type and gearing could be reused in the Logger application to reduce some of the math and make user's lives easier.

The units per rotation would have to be changed to something like (hopefully someone can come up with something shorter) "Post Gearing units per rotation".

Piphi5:
Fixes #288

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Logger.cpp
+++ b/sysid-application/src/main/native/cpp/view/Logger.cpp
@@ -106,11 +106,11 @@ void Logger::Display() {
                      m_isRotationalUnits ? ImGuiInputTextFlags_ReadOnly
                                          : ImGuiInputTextFlags_None);
   sysid::CreateTooltip(
-      "The logger assumes that the code will be sending recorded rotations "
-      "over NetworkTables. This value will then be multiplied by the units per "
-      "rotation to get the measurement in the units you specified.\n\nFor "
-      "non-rotational units (e.g. meters), this value is usually the wheel "
-      "diameter times pi.");
+      "The logger assumes that the code will be sending recorded output shaft "
+      "rotations over NetworkTables. This value will then be multiplied by the "
+      "units per rotation to get the measurement in the units you "
+      "specified.\n\nFor non-rotational units (e.g. meters), this value is "
+      "usually the wheel diameter times pi (should not include gearing).");
   // Create a section for voltage parameters.
   ImGui::Separator();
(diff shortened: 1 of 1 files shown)


sysid-application/src/main/native/cpp/view/Logger.cpp
```diff
@@ -106,11 +106,12 @@ void Logger::Display() {
                      m_isRotationalUnits ? ImGuiInputTextFlags_ReadOnly
                                          : ImGuiInputTextFlags_None);
   sysid::CreateTooltip(
-      "The logger assumes that the code will be sending recorded rotations "
+      "The logger assumes that the code will be sending recorded output shaft "
```
calcmogul, line 109: Here's the properly reflowed text:
```cpp
  sysid::CreateTooltip(
      "The logger assumes that the code will be sending recorded output shaft "
      "rotations over NetworkTables. This value will then be multiplied by the "
      "units per rotation to get the measurement in the units you "
      "specified.\n\nFor non-rotational units (e.g. meters), this value is "
      "usually the wheel diameter times pi (does not include gearing).");
```

sysid-application/src/main/native/cpp/view/Logger.cpp
```diff
@@ -106,11 +106,12 @@ void Logger::Display() {
...
+      "rotations "
       "over NetworkTables. This value will then be multiplied by the units per "
       "rotation to get the measurement in the units you specified.\n\nFor "
       "non-rotational units (e.g. meters), this value is usually the wheel "
-      "diameter times pi.");
+      "diameter times pi (does not include gearing).");
```
calcmogul, line 114: Maybe should not include gearing is a better phrase since saying does not leaves room for assuming maybe it should.

Piphi5: How so?

calcmogul: Why bring up gearing at all?

Piphi5: Ah that makes sense. The intent was to clarify the gearing thing because we did say the contrary on a Chief Delphi post. But I can see how that can make things worse for other teams.

calcmogul (approved):

PR #305, opened 2022-01-11
Change to acceleration-based OLS for all fits

Oblarg:
Acceleration is noisy, but if we place it as the dependent variable the noise does not matter nearly as much.  This lets us use the same fit for all of our analysis cases.

I've added a coefficient of determination for velocity computed from the sim data, as well.

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/FeedforwardAnalysis.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/FeedforwardAnalysis.cpp
@@ -16,52 +16,19 @@
 using namespace sysid;
 
 /**
- * Populates OLS vector for u = Ks + Kv v + Ka a.
+ * Populates OLS vector for x_k+1 - x_k / tau = alpha x_k + beta u_k + gamma
+ * sgn(x_k).
  *
  * @param d       List of characterization data.
  * @param type    Type of system being identified.
  * @param olsData Vector of OLS data.
  */
-static void PopulateAccelOLSVector(const std::vector<PreparedData>& d,
-                                   const AnalysisType& type,
-                                   std::vector<double>& olsData) {
+static void PopulateOLSVector(const std::vector<PreparedData>& d,
+                              const AnalysisType& type,
+                              std::vector<double>& olsData) {
   for (const auto& pt : d) {
-    // u = Ks sgn(v) + K_v v + Ka a
-
-    // Add the dependent variable (voltage)
-    olsData.push_back(pt.voltage);
-
-    // Add the intercept term (for Ks)
-    olsData.push_back(std::copysign(1, pt.velocity));
-
-    // Add the velocity term (for Kv)
-    olsData.push_back(pt.velocity);
-
-    // Add the acceleration term (for Ka)
+    // Add the dependent variable (acceleration)
     olsData.push_back(pt.acceleration);
-
-    if (type == analysis::kArm) {
-      // Add the cosine term (for Kcos)
-      olsData.push_back(pt.cos);
-    }
-  }
-}
-
-/**
- * Populates OLS vector for x_k+1 = alpha x_k + beta u_k + gamma sgn(x_k).
- *
- * @param d       List of characterization data.
- * @param type    Type of system being identified.
- * @param olsData Vector of OLS data.
- */
-static void PopulateNextVelOLSVector(const std::vector<PreparedData>& d,
-                                     const AnalysisType& type,
-                                     std::vector<double>& olsData) {
-  for (const auto& pt : d) {
-    // x_k+1 = alpha x_k + beta u_k + gamma sgn(x_k)
-
-    // Add the dependent variable (next velocity)
-    olsData.push_back(pt.nextVelocity);
 
     // Add the velocity term (for alpha)
     olsData.push_back(pt.velocity);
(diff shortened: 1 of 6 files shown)


sysid-application/src/main/native/cpp/analysis/FeedforwardAnalysis.cpp
```diff
@@ -86,41 +56,33 @@ std::tuple<std::vector<double>, double> sysid::CalculateFeedforwardGains(
...
-    PopulateAccelOLSVector(slow, type, olsData);
-    PopulateAccelOLSVector(fast, type, olsData);
 
-    // Gains are Ks, Kv, Ka, Kcos
-    return sysid::OLS(olsData, type.independentVariables);
-  } else {
-    // This implements the OLS algorithm defined in
-    // https://file.tavsys.net/control/sysid-ols.pdf.
+  olsData.reserve((2 + type.independentVariables) *
+                  (data.slow.size() - 1 + data.fast.size() - 1));
```
calcmogul, line 64: Document what these magic numbers correspond to.

sysid-application/src/main/native/cpp/analysis/FeedforwardAnalysis.cpp
```diff
@@ -86,41 +56,33 @@ std::tuple<std::vector<double>, double> sysid::CalculateFeedforwardGains(
...
 
-    // Initialize gains list with Ks, Kv, and Ka
-    std::vector<double> gains{
-        {-gamma / beta, (1 - alpha) / beta,
-         dtMean.value() * (alpha - 1) / (beta * std::log(alpha))}};
+  // Initialize gains list with Ks, Kv, and Ka
+  // std::vector<double> gains{
+  //     {-gamma / beta, (1 - alpha) / beta,
+  //      dtMean.value() * (alpha - 1) / (beta * std::log(alpha))}};
+  std::vector<double> gains{-gamma / beta, -alpha / beta, 1 / beta};
```
calcmogul, line 78: Add a derivation for these.

calcmogul (changes requested):

Piphi5:
Why is this better than the previous method that let us bypass acceleration noise entirely? 

Also do you have graphs to show the improvements?

Oblarg:
The other method is less robust to variation in timesteps (the timestep features explicitly in the conversions for the gains); the improvement is modest, but this method has consistently-lower simulated RMSE.

Oblarg:
re: performance

new algo: 
image

old algo:
image

Results are consistent across the 3 data sets I have access to.  Gains range from modest to nearly indistinguishable.  The sim R^2 for the new algo pretty clearly gives a better notion of goodness-of-fit than the raw OLS R^2 from the previous algorithm.

Oblarg:
I've deleted some tests that do not do anything, per the discussion on the linked issue.  The angular sim inserts the trackwidth into the gains for the underlying model, and then just divides it back out again when reading out the position/velocity.

sysid-application/src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -327,6 +330,7 @@ void AnalyzerPlot::SetData(const Storage& rawData, const Storage& filteredData,
   // represents the prediction at the timestep, and N represents the number of
   // points
   m_RMSE = std::sqrt(simSquaredErrorSum / timeSeriesPoints);
+  m_RSquared = 1 - m_RMSE / std::sqrt(squaredVariationSum / timeSeriesPoints);
```
calcmogul, line 333: This sets m_Rsquared to NaN if squaredVariationSum == 0.0 instead of the expected value of 1.

calcmogul (changes requested):

sysid-application/src/main/native/cpp/analysis/OLS.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "sysid/analysis/OLS.h"
 
+#include <iostream>
+#include <ostream>
```
calcmogul, line 8: Are these used?

sysid-application/src/main/native/cpp/view/AnalyzerPlot.cpp
Oblarg: This never actually occurs (it would require the data to be uniformly zero), and if it did the value of `1` isn't really helpful.  I don't think it's worth catching.

sysid-application/src/main/native/cpp/analysis/OLS.cpp
Oblarg: nope!

calcmogul:
@Piphi5 It looks like the JSON data upload broke. The data isn't in the jsons folder like it expected: https://github.com/wpilibsuite/sysid/runs/4806012450?check_suite_focus=true

sysid-application/src/main/native/cpp/analysis/FeedforwardAnalysis.cpp
```diff
@@ -76,51 +43,46 @@ static void PopulateNextVelOLSVector(const std::vector<PreparedData>& d,
...
+  // Perform OLS with accel = alpha*vel + beta*voltage + gamma*signum(vel)
+  // OLS performs best with the noisiest variable as the dependent var,
+  // so we regress accel in terms of the other variables.
+  PopulateOLSVector(slow, type, olsData);
+  PopulateOLSVector(fast, type, olsData);
+
+  auto ols = sysid::OLS(olsData, type.independentVariables);
+  double alpha = std::get<0>(ols)[0];  // -kv/ka
+  double beta = std::get<0>(ols)[1];   // 1/ka
+  double gamma = std::get<0>(ols)[2];  // ks/ka
```
calcmogul, line 75:
```suggestion
  double gamma = std::get<0>(ols)[2];  // -ks/ka
```

sysid-application/src/main/native/cpp/analysis/FeedforwardAnalysis.cpp
```diff
@@ -76,51 +43,46 @@ static void PopulateNextVelOLSVector(const std::vector<PreparedData>& d,
...
+
+  // Initialize gains list with Ks, Kv, and Ka
+  std::vector<double> gains{-gamma / beta, -alpha / beta, 1 / beta};
+
+  if (type == analysis::kElevator || type == analysis::kArm) {
+    // Add Kg to gains list
+    double delta = std::get<0>(ols)[3];  // kg/ka
```
calcmogul, line 82:
```suggestion
    double delta = std::get<0>(ols)[3];  // -kg/ka
```

calcmogul (changes requested):
Some negative signs were missing from the derivations for them to match up with the gain calculations.

Piphi5:
> @Piphi5 It looks like the JSON data upload on failure broke. The data isn't in the jsons folder like it expected: https://github.com/wpilibsuite/sysid/runs/4806012450?check_suite_focus=true
> 
> UPDATE: Apparently it happened because the quasistatic data trim removed all data.

from the source code it looks like right now it doesn't run for runtime errors

calcmogul (approved):

PR #306, opened 2022-01-12
Create proper SparkMax Presets

### LINKED ISSUE #258: Support FeedbackControllerPresets based on connected Encoder
https://github.com/wpilibsuite/sysid/pull/253#discussion_r771232288

Currently REV's brushless controller internal encoder uses a hall effect sensor with custom, unchangeable settings:

image

However, using the brushless mode on the motor controller apparently does not require to use the internal encoder, so using that setting may sometimes be wrong for the gains, hence it will be necessary to change the FeedbackControllerPreset based on the connected encoder.

See also, #259

Piphi5:
Fixes #258

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -38,8 +38,9 @@ Analyzer::Analyzer(glass::Storage& storage, wpi::Logger& logger)
   m_presets["WPILib (Pre-2020)"] = presets::kWPILibOld;
   m_presets["CANCoder"] = presets::kCTRECANCoder;
   m_presets["CTRE"] = presets::kCTREDefault;
-  m_presets["REV (Brushless)"] = presets::kREVBrushless;
-  m_presets["REV (Brushed)"] = presets::kREVBrushed;
+  m_presets["REV Brushless Encoder Port"] = presets::kREVNEOBuiltIn;
+  m_presets["REV Brushed Encoder Port"] = presets::kREVNonNEO;
+  m_presets["REV Data Port"] = presets::kREVNonNEO;
   m_presets["Venom"] = presets::kVenom;
 }
 
--- a/sysid-application/src/main/native/include/sysid/analysis/FeedbackControllerPreset.h
+++ b/sysid-application/src/main/native/include/sysid/analysis/FeedbackControllerPreset.h
@@ -123,6 +123,16 @@ constexpr FeedbackControllerPreset kCTREDefault{1023.0 / 12.0, 0.1, 1_ms, false,
                                                 81.5_ms};
 
 /**
+ * https://github.com/wpilibsuite/sysid/issues/258#issuecomment-1010658237
+ *
+ * 8-sample moving average with 32 ms between samples.
+ *
+ * Total delay = 8-tap moving average delay = (8 - 1) / 2 * 32 ms = 112 ms.
+ */
+constexpr FeedbackControllerPreset kREVNEOBuiltIn{1.0 / 12.0, 60.0, 1_ms, false,
+                                                  112_ms};
+
+/**
  * https://www.revrobotics.com/content/sw/max/sw-docs/cpp/classrev_1_1_c_a_n_encoder.html#a7e6ce792bc0c0558fb944771df572e6a
  *
  * Backward finite difference delay = 100 ms / 2 = 50 ms.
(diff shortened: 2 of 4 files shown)


sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -38,8 +38,8 @@ Analyzer::Analyzer(glass::Storage& storage, wpi::Logger& logger)
   m_presets["WPILib (Pre-2020)"] = presets::kWPILibOld;
   m_presets["CANCoder"] = presets::kCTRECANCoder;
   m_presets["CTRE"] = presets::kCTREDefault;
-  m_presets["REV (Brushless)"] = presets::kREVBrushless;
-  m_presets["REV (Brushed)"] = presets::kREVBrushed;
+  m_presets["REV (NEO Built-in Encoder)"] = presets::kREVNEOBuiltIn;
+  m_presets["REV (Non NEO Encoder)"] = presets::kREVNonNEO;
```
calcmogul, line 42: I think REV's docs call it "alternate encoder mode", so the string would be "REV (Alternate Encoder)".

sysid-application/src/main/native/include/sysid/analysis/FeedbackControllerPreset.h
```diff
@@ -122,6 +122,16 @@ constexpr FeedbackControllerPreset kCTRECANCoder{1.0 / 12.0, 60.0, 1_ms, true,
 constexpr FeedbackControllerPreset kCTREDefault{1023.0 / 12.0, 0.1, 1_ms, false,
                                                 81.5_ms};
 
+/**
+ * https://github.com/wpilibsuite/sysid/issues/258#issuecomment-1010658237
+ *
+ * 8 samples , 32 ms / sample
+ *
+ * Total delay = (8 - 1) / 2 * 32 = 112 ms
```
calcmogul, line 130:
```suggestion
 * Total delay = 8-tap moving average delay = (8 - 1) / 2 * 32 ms = 112 ms.
```

sysid-application/src/main/native/include/sysid/analysis/FeedbackControllerPreset.h
```diff
@@ -122,6 +122,16 @@ constexpr FeedbackControllerPreset kCTRECANCoder{1.0 / 12.0, 60.0, 1_ms, true,
 constexpr FeedbackControllerPreset kCTREDefault{1023.0 / 12.0, 0.1, 1_ms, false,
                                                 81.5_ms};
 
+/**
+ * https://github.com/wpilibsuite/sysid/issues/258#issuecomment-1010658237
+ *
+ * 8 samples , 32 ms / sample
```
calcmogul, line 128:
```suggestion
 * 8-sample moving average with 32 ms between samples.
```

calcmogul (changes requested):

sysid-application/src/main/native/cpp/view/Analyzer.cpp
Piphi5: But you could still use an Encoder on the Encoder port that technically isn't an Alternate encoder (iirc Alternate is for data port encoders) so I thought NEO vs Non NEO would make the difference more clear.

calcmogul: The encoder port is only intended for the communitation encoder, isn't it? The firmware wouldn't know whether a different motor was plugged into the encoder port, so it would apply the same filtering to whatever's plugged in.

Piphi5: The thing is that we support a Quadrature Encoder (for brushed mode) plugged into the Encoder port, so that's where that naming could get confusing. Though I do agree that doing NEO vs Non NEO is a point of confusion for the 550.

calcmogul: What about "REV (Brushless Built-in Encoder)", "REV (Brushed Encoder Port)", and "REV (Data Port)"?

Piphi5: Yeah that works

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -38,8 +38,9 @@ Analyzer::Analyzer(glass::Storage& storage, wpi::Logger& logger)
   m_presets["WPILib (Pre-2020)"] = presets::kWPILibOld;
   m_presets["CANCoder"] = presets::kCTRECANCoder;
   m_presets["CTRE"] = presets::kCTREDefault;
-  m_presets["REV (Brushless)"] = presets::kREVBrushless;
-  m_presets["REV (Brushed)"] = presets::kREVBrushed;
+  m_presets["REV Brushless Encoder Port"] = presets::kREVNEOBuiltIn;
+  m_presets["REV Brushed Encoder Port"] = presets::kREVNonNEO;
+  m_presets["REV Brushed Data Port"] = presets::kREVNonNEO;
```
calcmogul, line 43: The data port can be used with both brushless and brushed motors.

calcmogul (changes requested):

calcmogul (approved):

PR #308, opened 2022-01-15
Add JSON upload for analysis exceptions

Piphi5:
Uploads the data json if it catches a runtime error. This is an issue that was found in [this pr](https://github.com/wpilibsuite/sysid/pull/305#issuecomment-1012324368)

Made sure it worked by adding this line after [here](https://github.com/Piphi5/sysid/blob/integration-json-upload/sysid-application/src/integrationtest-analysis/native/cpp/AnalysisTest.cpp#L106) to trigger the quasistatic trim exception:
```
analyzerSettings.motionThreshold = 10000;
 ```

### DIFF
--- a/sysid-application/src/integrationtest-analysis/native/cpp/AnalysisTest.cpp
+++ b/sysid-application/src/integrationtest-analysis/native/cpp/AnalysisTest.cpp
@@ -61,6 +61,22 @@ class AnalysisTest : public ::testing::Test {
     Connect(m_nt, m_kill);
   }
 
+  void UploadJSON(std::string_view path) {
+    std::string jsonFolderPath =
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by calcmogul — APPROVED** [review]

File: sysid-application/src/integrationtest-analysis/native/cpp/AnalysisTest.cpp [hunk 1]
```diff
@@ -11,6 +11,7 @@
 #include <ntcore_c.h>
 #include <ntcore_cpp.h>
 #include <wpi/Logger.h>
+#include <wpi/fs.h>
```
calcmogul, line 14: Was this actually used?

calcmogul (approved):

PR #309, opened 2022-01-15
Add angular drivetrain unit tests for feedforward analysis

### LINKED ISSUE #307: Fix Angular Drivetrain Unit Tests
Currently the unit tests fail to test trackwidth:
```
double AngularDriveSim::GetPosition() const {
  return m_x(0) * m_trackwidth / 2;
}

double AngularDriveSim::GetVelocity() const {
  // The below comes from v = ω * r.
  return m_x(1) * m_trackwidth / 2;
}
```

It's also the reason behind #305 failing

calcmogul:
SimpleMotorSim is used to generate the data because we only care about
coverage of the different code paths in CalculateFeedforwardGains(), not
where the data used by the fit comes from. The analysis integration
tests cover logging for a realistic drivetrain model.

Fixes #307.


Piphi5 (approved):

Oblarg:
This does test the different code path, but the analysis is identical - why are these separate options?

calcmogul:
Because it determines whether the trackwidth calc gets run elsewhere.

Oblarg:
Ah, that makes sense.

Oblarg (approved):
LGTM

Issue #312, opened 2022-01-17
ImGUI crashes on JSON with data below motion threshold


calcmogul:
Here's the JSON.
[sysid_data20220117-182153.json.txt](https://github.com/wpilibsuite/sysid/files/7883734/sysid_data20220117-182153.json.txt)

The first run opens a dialog saying the quasistatic test had all the data trimmed. Here's the logs.
```
> Task :sysid-application:run
INFO: Opening File: /tmp/sysid_data20220117-182153.json (Analyzer.cpp:490)
INFO: Opened File (Analyzer.cpp:494)
INFO: Read /tmp/sysid_data20220117-182153.json (AnalysisManager.cpp:573)
INFO: Parsing initial data of /tmp/sysid_data20220117-182153.json (AnalysisManager.cpp:594)
INFO: Preparing Data. (Analyzer.cpp:510)
INFO: Preparing Drivetrain data (AnalysisManager.cpp:608)
INFO: Reading JSON data. (AnalysisManager.cpp:431)
INFO: Preprocessing raw data. (AnalysisManager.cpp:438)
INFO: Copying raw data. (AnalysisManager.cpp:450)
INFO: Converting to PreparedData struct. (AnalysisManager.cpp:454)
INFO: Applying trimming and filtering. (AnalysisManager.cpp:501)
INFO: Calculating Gains. (Analyzer.cpp:525)
INFO: Returning early for gain calculation. (Analyzer.cpp:527)
INFO: Graphing Data. (Analyzer.cpp:552)
INFO: Attempting to graph only raw time series. (Analyzer.cpp:556)
```
When SysId is closed and reopened, it crashes with the following exception from ImGUI.
```
> Task :sysid-application:run
INFO: Previous file location exists: /tmp/sysid_data20220117-182153.json (Analyzer.cpp:81)
INFO: Read /tmp/sysid_data20220117-182153.json (AnalysisManager.cpp:573)
INFO: Parsing initial data of /tmp/sysid_data20220117-182153.json (AnalysisManager.cpp:594)
INFO: Preparing Data. (Analyzer.cpp:510)
INFO: Preparing Drivetrain data (AnalysisManager.cpp:608)
INFO: Reading JSON data. (AnalysisManager.cpp:431)
INFO: Preprocessing raw data. (AnalysisManager.cpp:438)
INFO: Copying raw data. (AnalysisManager.cpp:450)
INFO: Converting to PreparedData struct. (AnalysisManager.cpp:454)
[...]

PR #314, opened 2022-01-19
Remove loading previous JSON files

### LINKED ISSUE #313: Don't load the previous JSON on startup
Developers do this during testing, but it can get annoying when working with malformed JSONs. Also, loading the previous JSON automatically doesn't fit a typical user's workflow. The user will usually be selecting a different one via the file selection UI.

Piphi5:
Fixes #313

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -598,6 +598,8 @@ AnalysisManager::AnalysisManager(std::string_view path, Settings& settings,
   // Get the rotation -> output units factor from the JSON.
   m_unit = m_json.at("units").get<std::string>();
   m_factor = m_json.at("unitsPerRotation").get<double>();
+  WPI_DEBUG(m_logger, "Parsing units per rotation as {} {} per rotation",
+            m_factor, m_unit);
 
   // Reset settings for Dynamic Test Limits
   m_settings.stepTestDuration = units::second_t{0.0};
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -31,7 +31,7 @@ using namespace sysid;
 static double kBadDataGain = -9999.0;
 
 Analyzer::Analyzer(glass::Storage& storage, wpi::Logger& logger)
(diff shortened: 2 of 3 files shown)


calcmogul (approved):
Works locally for me on Linux.

Piphi5:
@calcmogul do you think this also solves the crashing issue with #312

calcmogul:
I just tested it. It does. However, what's the deal with the units per rotation?
Screenshot_20220119_091818

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -598,6 +598,7 @@ AnalysisManager::AnalysisManager(std::string_view path, Settings& settings,
   // Get the rotation -> output units factor from the JSON.
   m_unit = m_json.at("units").get<std::string>();
   m_factor = m_json.at("unitsPerRotation").get<double>();
+  WPI_DEBUG(m_logger, "Parsing units per rotation as {}, {}", m_unit, m_factor);
```
calcmogul, line 601:
```suggestion
  WPI_DEBUG(m_logger, "Parsing units per rotation as {} {} per rotation", m_factor, m_units);
```
This will make it print `Parsing units per rotation as 0.2 Meters per rotation`.

calcmogul (changes requested):

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -498,6 +460,8 @@ void Analyzer::SelectFile() {
       m_manager =
           std::make_unique<AnalysisManager>(m_location, m_settings, m_logger);
       m_type = m_manager->GetAnalysisType();
+      m_factor = m_manager->GetFactor();
+      m_unit = m_manager->GetUnit();
```
calcmogul, line 464: Just noting that this fixes a bug where units per rotation wasn't initialized if the JSON failed to load, so it contained garbage.

calcmogul (approved):

caseyjbrotherton:
Just as an FYI:
 looks like the root cause is https://github.com/ocornut/imgui/issues/4920.

Whatever error is displayed from the json causes the popup.  And the popup before a window is created causes the assert.
( removing the preload is a fine workaround.  If you end up wanting the preload, then you may be able to look at this fix. )

PR #316, opened 2022-01-22
Fix bugs in generation UI's gyro parameter

calcmogul:
Make gyro parameter in generation UI more descriptive. Instead of saying
"Gyro Parameter", it's now more specific (e.g., "CAN ID", "SPI Port",
"Analog Port".

A combo box was added for the ADIS IMU SPI port. MXP is first for the
ADIS16448 and onboard CS0 is first for the ADIS16470 since those are the
most common SPI ports they're plugged into respectively.

The parameter options were rewritten to be easier to read since
formatting them like enums isn't required.

The "no gyro" option now no longer displays a gyro parameter input box.

The generation test was using the ADXRS450 constructor arguments for the
ADIS16448, which was wrong.

### DIFF
--- a/sysid-library/src/main/cpp/generation/SysIdSetup.cpp
+++ b/sysid-library/src/main/cpp/generation/SysIdSetup.cpp
@@ -61,13 +61,13 @@ void AddMotorController(
   if (controller == "TalonSRX" || controller == "VictorSPX" ||
       controller == "TalonFX") {
     if (controller == "TalonSRX") {
-      fmt::print("Setup TalonSRX");
+      fmt::print("Setup TalonSRX\n");
       controllers->push_back(std::make_unique<WPI_TalonSRX>(port));
     } else if (controller == "TalonFX") {
-      fmt::print("Setup TalonFX");
+      fmt::print("Setup TalonFX\n");
       controllers->emplace_back(std::make_unique<WPI_TalonFX>(port));
     } else {
-      fmt::print("Setup VictorSPX");
+      fmt::print("Setup VictorSPX\n");
       controllers->emplace_back(std::make_unique<WPI_VictorSPX>(port));
     }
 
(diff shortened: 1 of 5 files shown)


sysid-library/src/main/cpp/generation/SysIdSetup.cpp
```diff
@@ -389,49 +375,62 @@ void SetupGyro(
...
-      fmt::print("Setup Romi");
+      fmt::print("Setup Romi\n");
 #ifndef __FRC_ROBORIO__
       gyro = std::make_unique<frc::RomiGyro>();
 #endif
-    } else {
+    } else if (gyroType == "Analog Gyro") {
       try {
-        fmt::print("Setup Analog Gyro, Port: {}", gyroCtor);
+        fmt::print("Setup Analog Gyro, Analog Port {}\n", gyroCtor);
```
Piphi5, line 429: This doesn't match the unit test search

sysid-library/src/main/cpp/generation/SysIdSetup.cpp
```diff
@@ -266,20 +266,20 @@ void SetupEncoders(
...
     return frc::SPI::Port::kOnboardCS2;
-  } else if (name == "kOnboardCS3") {
-    fmt::print("Setup SPI kOnboardCS3");
+  } else if (name == "SPI (Onboard CS3)") {
+    fmt::print("Setup SPI (Onboard CS3)\n");
     return frc::SPI::Port::kOnboardCS3;
-  } else if (name == "SPI.kMXP") {
-    fmt::print("Setup SPI SPI.kMXP");
+  } else if (name == "SPI (MXP)") {
+    fmt::print("Setup SPI (MXP)\n");
```
Piphi5, line 282: Either we make a separate integration test clause for this or we make it match the default format

sysid-library/src/main/cpp/generation/SysIdSetup.cpp
```diff
@@ -358,25 +358,11 @@ void SetupGyro(
       units::degrees_per_second_t rate{xyz_dps[2]};
       return units::radians_per_second_t{rate}.value();
     };
-    fmt::print("Setup Pigeon, Port: {}", portStr);
+    fmt::print("Setup Pigeon, Port: {}\n", portStr);
```
Piphi5, line 361: I think if you remove `Port:` you should be good

Piphi5 (changes requested):

Piphi5 (approved):
Works for me and looks great!

PR #320, opened 2022-01-23
Pass CANCoder Gearing by value

### LINKED ISSUE #318: Incorrect Position measurement using CANCoder
We were testing Sysid on our turret, which is a VictorSPX controlling a 775 motor with a CANCoder as its remote sensor. In our results, we saw that the position value returned to SmartDashboard was way off even when not running any of the 4 tests. The position was reporting a number like 4e+60 as our turret's current position. Checking the values the CANCoder returned using Phoenix Tuner showed that it was reporting a correct value when we manually turned our turret.  We are using the latest version of Phoenix Tuner with the latest CANCoder firmware (22.0.0.0). I have attached the configuration we used and the data that our team collected from running the characterization tool on our turret.

[turretconfig - Copy.txt](https://github.com/wpilibsuite/sysid/files/7919520/turretconfig.-.Copy.txt)
[sysid_data20220122-143225 - Copy.txt](https://github.com/wpilibsuite/sysid/files/7919521/sysid_data20220122-143225.-.Copy.txt)

Piphi5:
Fixes #318

### DIFF
--- a/sysid-library/src/main/cpp/generation/SysIdSetup.cpp
+++ b/sysid-library/src/main/cpp/generation/SysIdSetup.cpp
@@ -244,8 +244,8 @@ void SetupEncoders(
     cancoder->ConfigVelocityMeasurementPeriod(cancoderPeriod);
     cancoder->ConfigVelocityMeasurementWindow(numSamples);
 
-    position = [&] { return cancoder->GetPosition() / combinedCPR; };
-    rate = [&] { return cancoder->GetVelocity() / combinedCPR; };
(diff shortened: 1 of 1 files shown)


prateekma (approved):

Piphi5:
I think we should make sure it resolves the issue before merging

calcmogul:
Did it?

Piphi5:
From what I've gathered on #318 is that the CANCoder issue has been fixed but the person was running into other trouble with his units so the I'm not sure if the issue can be closed yet.

PR #330, opened 2022-01-26
Use numerical derivative to compute gyro angular rate

calcmogul:


calcmogul:
The angular velocity has too much banding due to low resolution angular position data. We're going to use encoders for the drivetrain angular test instead.

PR #332, opened 2022-01-26
Remove median filter

Oblarg:
Removes the median filter, as it no longer appreciably improves the quality of the fit on actual robot data (and in scenarios where it does, the sensor noise should probably be fixed anyway...)

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/FilteringUtils.cpp
@@ -10,7 +10,6 @@
 
 #include <fmt/format.h>
 #include <frc/filter/LinearFilter.h>
-#include <frc/filter/MedianFilter.h>
 #include <units/math.h>
 #include <wpi/StringExtras.h>
 #include <wpi/numbers>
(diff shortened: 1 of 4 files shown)


calcmogul (changes requested):
You forgot to remove the MedianFilter.h include from FilteringUtils.cpp.

calcmogul:
The fit on the following data is much worse without the median filter.
[sysid_data20220119-150715.json.txt](https://github.com/wpilibsuite/sysid/files/7943784/sysid_data20220119-150715.json.txt)

With median filter:
median-filter

Without median filter:
no-median-filter

If our stance is to fix the measurement noise, how would one do that? This data was generated by sysid.

Oblarg:
ooof, I had never seen data quite that bad.  Any idea what the setup is that's generating that?

calcmogul:
Unfortunately, I don't know what the hardware setup was. I didn't ask for their generation JSON.

Oblarg:
Okay, noted; I'll change the PR to simply defaulting the filter to off (i.e. window size 1), since it does dilute the acceleration gain if the noise is acceptable.

calcmogul:
Sounds good.

PR #333, opened 2022-01-26
Add characteristic timescale reporting, display tooltips always

Oblarg:
Reports the characteristic timescale of the mechanism, along with a tooltip that explains how to interpret it.

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -531,6 +531,7 @@ void Analyzer::Calculate() {
     const auto& [ff, fb, trackWidth] = m_manager->Calculate();
     m_ff = std::get<0>(ff);
     m_rSquared = std::get<1>(ff);
+    m_timescale = m_ff[2] / m_ff[1];
     m_Kp = fb.Kp;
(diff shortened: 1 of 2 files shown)


Piphi5:
Do the tooltips display nicely on the combined view?

Oblarg:
seems so

image

Piphi5 (approved):

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -628,35 +629,34 @@ void Analyzer::DisplayFeedforwardGains(bool combined) {
     DisplayGain("Track Width", &*m_trackWidth);
   }
 
+  SetPosition(beginX, beginY, 0, 3);
+  DisplayGain("Response Timescale (s)", &m_timescale);
+  CreateTooltip(
+      "The characteristic timescale of the system response in seconds. "
+      "Both the control loop period and total signal delay should be "
```
calcmogul, line 636: Time periods being faster makes no sense. You could either say the rates should be faster or the periods should be shorter.

calcmogul (changes requested):

calcmogul (approved):

PR #339, opened 2022-01-27
Refactor plots

### LINKED ISSUE #269: Refactor plot display code
Currently, there is a lot of [hard coding](https://github.com/wpilibsuite/sysid/pull/268#pullrequestreview-839996036) that makes it difficult to customize the order of the plots if need be.

Also, the code could probably be cleaned up alot

calcmogul:
The voltage domain plots were removed since they aren't relevant to the
fit anymore. The acceleration plots were removed since they are almost
always noise and aren't helpful for troubleshooting.

The GUI layout was reorganized so all the plots fit in the main display
and the only windows with a scrollbar are the generator and program log.

Instances of "units" are now replaced with an abbreviated unit name so
they use less horizontal space.

Fixes #269.

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -21,6 +21,7 @@
 #include <wpi/json.h>
 #include <wpi/raw_istream.h>
 
+#include "sysid/Util.h"
 #include "sysid/analysis/AnalysisType.h"
 #include "sysid/analysis/FilteringUtils.h"
 #include "sysid/analysis/JSONConverter.h"
@@ -595,6 +596,12 @@ AnalysisManager::AnalysisManager(std::string_view path, Settings& settings,
 
   // Get the rotation -> output units factor from the JSON.
   m_unit = m_json.at("units").get<std::string>();
+  for (int i = 0; i < sizeof(kUnits); ++i) {
+    if (m_unit == kUnits[i]) {
+      m_abbrevUnit = GetAbbreviation(kUnits[i]);
+      break;
+    }
+  }
   m_factor = m_json.at("unitsPerRotation").get<double>();
   WPI_DEBUG(m_logger, "Parsing units per rotation as {} {} per rotation",
             m_factor, m_unit);
@@ -652,14 +659,22 @@ AnalysisManager::Gains AnalysisManager::Calculate() {
 }
 
 void AnalysisManager::OverrideUnits(std::string_view unit,
+                                    std::string_view abbrevUnit,
                                     double unitsPerRotation) {
   m_unit = unit;
+  m_abbrevUnit = abbrevUnit;
   m_factor = unitsPerRotation;
   PrepareData();
 }
 
 void AnalysisManager::ResetUnitsFromJSON() {
   m_unit = m_json.at("units").get<std::string>();
+  for (int i = 0; i < sizeof(kUnits); ++i) {
+    if (m_unit == kUnits[i]) {
+      m_abbrevUnit = GetAbbreviation(kUnits[i]);
+      break;
+    }
+  }
   m_factor = m_json.at("unitsPerRotation").get<double>();
   PrepareData();
 }
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -626,6 +626,14 @@ void AnalysisManager::PrepareData() {
 }
 
 AnalysisManager::Gains AnalysisManager::Calculate() {
+  if (m_filteredDatasets.empty()) {
+    throw std::runtime_error(
+        "There is an unresolved issue with the data being used. Please "
+        "double-check the data quality and adjust settings such as units, "
+        "velocity threshold, test duration, etc. to make sure that there are "
+        "no errors with the dataset before attempting to re-calculate gains.");
+  }
+
   WPI_INFO(m_logger, "{}", "Calculating Gains");
   // Calculate feedforward gains from the data.
   auto ffGains = sysid::CalculateFeedforwardGains(
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -596,12 +596,6 @@ AnalysisManager::AnalysisManager(std::string_view path, Settings& settings,
 
   // Get the rotation -> output units factor from the JSON.
   m_unit = m_json.at("units").get<std::string>();
-  for (int i = 0; i < sizeof(kUnits); ++i) {
-    if (m_unit == kUnits[i]) {
-      m_abbrevUnit = GetAbbreviation(kUnits[i]);
-      break;
-    }
-  }
   m_factor = m_json.at("unitsPerRotation").get<double>();
   WPI_DEBUG(m_logger, "Parsing units per rotation as {} {} per rotation",
             m_factor, m_unit);
@@ -659,22 +653,14 @@ AnalysisManager::Gains AnalysisManager::Calculate() {
 }
 
 void AnalysisManager::OverrideUnits(std::string_view unit,
-                                    std::string_view abbrevUnit,
                                     double unitsPerRotation) {
   m_unit = unit;
-  m_abbrevUnit = abbrevUnit;
   m_factor = unitsPerRotation;
   PrepareData();
 }
 
 void AnalysisManager::ResetUnitsFromJSON() {
   m_unit = m_json.at("units").get<std::string>();
-  for (int i = 0; i < sizeof(kUnits); ++i) {
-    if (m_unit == kUnits[i]) {
-      m_abbrevUnit = GetAbbreviation(kUnits[i]);
-      break;
-    }
-  }
   m_factor = m_json.at("unitsPerRotation").get<double>();
   PrepareData();
 }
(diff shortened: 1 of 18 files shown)


calcmogul:
Screenshot_20220126_193422

Oblarg:
Think you can add a velocity vs acceleration plot to fill out the plot panel?  That gives a good summary of linearity of the data.

calcmogul:
I still need to refactor the plotting code more to remove duplication and make it easier to add or remove specific plots, but here's what's implemented so far.
[sysid_data20220119-150715.json.txt](https://github.com/wpilibsuite/sysid/files/7956633/sysid_data20220119-150715.json.txt)
Screenshot_20220127_232339

Oblarg:
I'd say the "attempt" worked really really well!  That's a very nice set of summary plots.

sysid-application/src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -284,13 +218,85 @@ void AnalyzerPlot::SetData(const Storage& rawData, const Storage& filteredData,
...
+  auto maxVel = std::max(slowMaxVel, fastMaxVel);
+  m_stateGainFit[1] = ImPlotPoint(maxVel, -Kv / Ka * maxVel);
+
+  // Populate acceleration vs velocity graph
+  for (size_t i = 0; i < slow.size(); i += slowStep) {
+    if (abort) {
+      return;
+    }
+
+    // Calculate portion of voltage that corresponds to change in acceleration.
```
Oblarg, line 264: should be "portion of acceleration caused by back-EMF"

calcmogul:
I finished the AnalyzerPlot refactor. It was almost a rewrite.

Oblarg (dismissed):
LGTM!  Huge improvement.

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -595,6 +596,12 @@ AnalysisManager::AnalysisManager(std::string_view path, Settings& settings,
 
   // Get the rotation -> output units factor from the JSON.
   m_unit = m_json.at("units").get<std::string>();
+  for (int i = 0; i < sizeof(kUnits); ++i) {
+    if (m_unit == kUnits[i]) {
+      m_abbrevUnit = GetAbbreviation(kUnits[i]);
+      break;
+    }
+  }
   m_factor = m_json.at("unitsPerRotation").get<double>();
```
Piphi5, line 605: It might be worth reducing a little bit of code duplication by using the `ResetUnitsFromJSON` function here. The function would need a better name though

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -652,14 +659,22 @@ AnalysisManager::Gains AnalysisManager::Calculate() {
 }
 
 void AnalysisManager::OverrideUnits(std::string_view unit,
+                                    std::string_view abbrevUnit,
                                     double unitsPerRotation) {
   m_unit = unit;
+  m_abbrevUnit = abbrevUnit;
```
Piphi5, line 665: Shouldn't we handle figuring out abbrevUnit in this function?

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -87,7 +89,7 @@ void Analyzer::Display() {
   // Allow the user to select which data set they want analyzed and add a reset
   // button. Also show the units and the units per rotation.
   if (m_manager) {
-    ImGui::SetNextItemWidth(ImGui::GetFontSize() * 15);
+    ImGui::SetNextItemWidth(ImGui::GetFontSize() * 10);
```
Piphi5, line 92: See comment about using 10 as a constant.

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -595,6 +596,12 @@ AnalysisManager::AnalysisManager(std::string_view path, Settings& settings,
 
   // Get the rotation -> output units factor from the JSON.
   m_unit = m_json.at("units").get<std::string>();
+  for (int i = 0; i < sizeof(kUnits); ++i) {
+    if (m_unit == kUnits[i]) {
+      m_abbrevUnit = GetAbbreviation(kUnits[i]);
+      break;
+    }
+  }
```
Piphi5, line 604: Why not just use GetAbbreviation to start with?

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
Piphi5, line 665: Shouldn't we handle the abbreviated Unit here within the function itself?

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -661,36 +436,217 @@ void Analyzer::DisplayFeedforwardGains(bool combined) {
...
 
-  SetPosition(beginX, beginY, horizontalSpacing, combined ? 2 : 3);
+  SetPosition(beginX, beginY, kHorizontalOffset, 2);
+  ImGui::SetNextItemWidth(ImGui::GetFontSize() * 4);
+  if (ImGui::SliderFloat("Test Duration", &m_stepTestDuration,
+                         m_manager->GetMinDuration(),
+                         m_manager->GetMaxDuration(), "%.2f")) {
+    m_settings.stepTestDuration = units::second_t{m_stepTestDuration};
+    m_enabled = true;
+    RefreshInformation();
```
Piphi5, line 458: Should we also make this another function?

sysid-application/src/main/native/cpp/view/AnalyzerPlot.cpp
```diff
@@ -7,38 +7,29 @@
...
-static int timeSeriesPoints = 0;
-
 template <typename Model>
 static std::vector<std::vector<ImPlotPoint>> PopulateTimeDomainSim(
     const std::vector<PreparedData>& data,
-    const std::array<units::second_t, 4>& startTimes, size_t step,
-    Model model) {
+    const std::array<units::second_t, 4>& startTimes, size_t step, Model model,
+    double& simSquaredErrorSum, double& squaredVariationSum,
+    int& timeSeriesPoints) {
```
Piphi5, line 32: If we're modifying variables, shouldn't we be using pointers or does that not apply to primitives?

Piphi5 (changes requested):

Piphi5:
A few minor cleanup issues, but overall I think the setup + refactored code works great.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -661,36 +436,217 @@ void Analyzer::DisplayFeedforwardGains(bool combined) {
...
-                           m_manager->GetMinDuration(),
-                           m_manager->GetMaxDuration(), "%.2f")) {
-      m_settings.stepTestDuration = units::second_t{m_stepTestDuration};
+  value = m_settings.preset.outputVelocityTimeFactor;
+  if (ImGui::InputDouble("Velocity Denominator Units (s)", &value, 0.0, 0.0,
+                         "%.1f") &&
+      value > 0) {
+    m_settings.preset.outputVelocityTimeFactor = value;
+    m_enabled = true;
+    Calculate();
```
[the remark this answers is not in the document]
calcmogul: I don't know enough about this code to do that kind of refactor without breaking it (I just spent an hour or so trying). It's not clear when enabled should be set to true and when it shouldn't, and the state of enabled sets reset at random times depending on when an exception is thrown.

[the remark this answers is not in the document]
Piphi5: Ideally enabled is set to false whenever an exception is thrown and it's set to true whenever someone updates a setting. I was just thinking we could put those two lines in another function like RecalculateGains() and replace any occurrences of these two lines with that function. Does that cause it to break?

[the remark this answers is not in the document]
calcmogul: When does one use Calculate() vs Recalculate()?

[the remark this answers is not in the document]
Piphi5: Calculate only reruns the Feedforward + Feedback Gain calculation while Recalculate reruns the data preparation (filtering, etc.) and then runs Calculate(). So the reason is mainly for optimization purposes (e.g. changing the measurement delay doesn't merit redoing the data preparation).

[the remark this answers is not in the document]
calcmogul: But why does setting enabled to true cause the data prep to rerun? That's not obvious at all looking at the call site of Calculate().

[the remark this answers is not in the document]
Piphi5: oh whoops I had a brainfart and explained the difference between Calculate and RefreshInformation()

Recalculate should only be used whenever the GUI changes a value as that change may fix the exception or issue that caused the GUI to disable.

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
calcmogul: Should PrepareData() be called in the constructor though?

Piphi5: Ah it probably shouldn't so I that should be fine especially with the updated changes

sysid-application/src/main/native/cpp/view/Analyzer.cpp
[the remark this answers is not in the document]
calcmogul: I think some of the functions are catching the exceptions too early. If we move the catch higher up in the call tree, we can probably remove a lot of the enabled checks.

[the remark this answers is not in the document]
Piphi5: Would that be something to fix in another PR?

Piphi5 (approved):

calcmogul:
No JSON:
sysid-no-json

Angular data quasistatic empty:
sysid-quasistatic-empty

Angular data success:
sysid-angular-success

Linear data success:
sysid-linear-success

PR #342, opened 2022-01-28
Add more robust analysis error handling

### LINKED ISSUE #340: Crash on windows 10 64-bit
Hey there, team 2718 here.
After running the tests to generate data for our drivetrain (although the tests were over a fairly short distance and time), we loaded it into the analyzer to analyze it. We then got a message that said "Quasistatic test trimming removed all data." So we were going to go and redo the tests to get better data, but for some reason one of the programmers decided to click the "dataset" drop-down box and select "forward," which causes the tool to crash. Upon reopening we got the assertion errors screenshotted below. The tool is now broken and will not work again, even after restarting the computer. The data set that caused this is also attached below. The tool does not break, however, after being installed directly from github instead of the FRC-2022 WPILib installer. It still crashes whenever "forward" is selected.

Capture
Capture1
[sysid_data20220127-190536.zip](https://github.com/wpilibsuite/sysid/files/7955020/sysid_data20220127-190536.zip)

Piphi5:
There is now repreparation of the data and exception handling if any part of the feedforward analyzer GUI section is updated.

For feedback calculations, an exception is now thrown if there isn't any data to calculate gains from. This error is subsequently handled in the GUI.

Fixes #340 (wrt the crashing issue)

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -625,6 +625,14 @@ void AnalysisManager::PrepareData() {
 }
 
 AnalysisManager::Gains AnalysisManager::Calculate() {
+  if (m_filteredDatasets.empty()) {
+    throw std::runtime_error(
+        "There is an unresolved issue with the data being used. Please "
+        "double-check the data quality and adjust settings such as units, "
+        "velocity threshold, test duration, etc. to make sure that there are "
+        "no errors with the dataset before attempting to re-calculate gains.");
+  }
+
   WPI_INFO(m_logger, "{}", "Calculating Gains");
   // Calculate feedforward gains from the data.
   auto ffGains = sysid::CalculateFeedforwardGains(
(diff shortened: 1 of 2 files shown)


sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -625,6 +625,10 @@ void AnalysisManager::PrepareData() {
 }
 
 AnalysisManager::Gains AnalysisManager::Calculate() {
+  if (m_filteredDatasets.empty()) {
+    throw std::runtime_error("Invalid data to perform gain calculation.");
```
calcmogul, line 629: What makes the data invalid and why did it happen? This error message isn't helpful.

calcmogul (changes requested):

sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
Piphi5: In essence, this will run if there was an error with the data analysis (e.g. the quasistatic trim) and the user manages to trigger the gain calculation function.

Would this work:
"There is an unresolved issue with the data being used. Please double-check the data quality and adjust settings such as units, velocity threshold, test duration, etc. to make sure that there are no errors with the dataset before attempting to calculate gains."

calcmogul: Yea.

calcmogul (approved):

PR #344, opened 2022-01-29
Upgrade to C++20 and NetworkTables 4

calcmogul:
This makes std::copy_n() (for concatenating std::arrays) and
fmt::format() format string parsing constexpr.


sysid-application/src/integrationtest-utils/native/cpp/IntegrationUtils.cpp
```diff
@@ -87,9 +86,9 @@ std::string KillNT(NT_Inst nt, NT_Entry kill) {
   fmt::print(stderr, "Killed robot program\n");
 
   // Set kill entry to false for future tests
-  nt::SetEntryValue(kill, nt::Value::MakeBoolean(false));
+  kill.Set(false);
   // Stop NT Client.
-  nt::StopClient(nt);
+  nt.StopClient();
```
PeterJohnson, line 91: Here you're setting kill to false and then immediately stopping the client.  This doesn't allow enough time for the value to actually propagate.

PR #345, opened 2022-01-31
Make plot point size configurable

Oblarg:
my elderly aching eyes can't pick out the dead-pixel-like isolated points under the current settings

### DIFF
--- a/sysid-application/src/main/native/include/sysid/view/AnalyzerPlot.h
+++ b/sysid-application/src/main/native/include/sysid/view/AnalyzerPlot.h
@@ -26,6 +26,8 @@ namespace sysid {
  */
 class AnalyzerPlot {
  public:
+  float m_pointSize = 1.25;
+
   /**
    * Constructs an instance of the analyzer plot helper and allocates memory for
    * all data vectors.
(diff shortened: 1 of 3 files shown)


Piphi5:
Can you send screenshots of the new plots?

calcmogul:
Suggestion from the FRC Discord: since smaller point sizes hide less information and accurately present noise, we'd want to support both sizes. We could add a point size slider to the diagnostic plots window with a range of 1 to 2 (1 by default). It would go in the same block of code as the R^2 so it's hidden when the plots aren't in use.

Oblarg:
image

calcmogul:
You should rebase. The plot resizing only provides room for three text boxes, so there will be a scroll bar.

Oblarg:
I just did rebase.

calcmogul:
Seems like the "available area" automatically decreased by putting a slider before the plots. Neat.

sysid-application/src/main/native/include/sysid/view/AnalyzerPlot.h
```diff
@@ -112,6 +112,8 @@ class AnalyzerPlot {
    */
   double* GetSimRSquared();
 
+  float m_pointSize = 1.25;
```
calcmogul, line 115: We generally put variables first, then functions in a given access specifier block.

calcmogul (changes requested):

calcmogul (approved):

PR #346, opened 2022-01-31
Make plots resize with their containing window

calcmogul:


calcmogul:
Plots with default size and without data:
plots-default-size-without-data

Plots with default size and with data:
plots-default-size-with-data

Plots resized and with data:
plots-resized-with-data

Piphi5 (approved):
Looks great

PR #347, opened 2022-02-01
Compute default motion threshold

### LINKED ISSUE #327: Choose a better default motion threshold for angular test
The angular rate seems to be way lower on average than 0.2 because robots don't rotate in yaw as fast as the drivetrain wheels rotate. We should pick a smaller default motion threshold for the angular test specifically.

Piphi5:
This should automatically calculate the velocity motion threshold for quasistatic tests through the noise floor from a moving standard deviation.

Fixes #327


calcmogul (approved):
It worked on the data I threw at it locally, including some angular drivetrain test data that used to throw an exception for having a motion threshold that's too high.

calcmogul (approved):

PR #349, opened 2022-02-01
Remove subsets of analysis datasets

### LINKED ISSUE #341: Remove dataset combo box
Users are more likely to get inaccurate fits from choosing different subsets of their data (e.g. forward, backward) and as a result, the dataset combo box in the Analyzer should be removed in addition to any analysis code made to store these different subsets.

For reference this is the combo box:
image

Piphi5:
Fixes #341

### DIFF
--- a/sysid-application/src/main/native/include/sysid/analysis/Storage.h
+++ b/sysid-application/src/main/native/include/sysid/analysis/Storage.h
@@ -77,6 +77,8 @@ struct Storage {
    * Dataset for fast (aka dynamic) test
    */
   std::vector<PreparedData> fast;
+
+  bool empty() const { return slow.empty() || fast.empty(); }
 };
 
 }  // namespace sysid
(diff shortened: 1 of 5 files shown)


calcmogul:
Looks like some files may be able to have the StringMap include removed.

sysid-application/src/main/native/include/sysid/analysis/Storage.h
```diff
@@ -77,6 +77,8 @@ struct Storage {
    * Dataset for fast (aka dynamic) test
    */
   std::vector<PreparedData> fast;
+
+  bool empty() const { return fast.empty() || slow.empty(); }
```
calcmogul, line 81: Please make the order of empty() calls match the member declaration order. :)

calcmogul:
The analyzer reset button is overlapping the file select box. Since it resets the whole analyzer, I guess it makes sense in the corner. The select box should just be shrunk a bit.
Screenshot_20220131_213244
Also, the analyzer window height could be reduced in UILayout.h to leave more room for the program log window. Just leave one text box worth of space for the trackwidth box in the angular drivetrain test.

Piphi5:
Here's a sample with elevator data 
image

calcmogul:
Looks like another 5 pixels or so would get rid of the analysis window scrollbar.

Piphi5:
Yeah its gone:
image

calcmogul (approved):

PR #351, opened 2022-02-02
Fix build with new FPGA image

calcmogul:


Piphi5:
Does this fix the bug with the gyros?

calcmogul:
Yes. Main builds won't work on real hardware until the new image is released tho.

PR #353, opened 2022-02-02
Separate forward/backward plots

Oblarg:
Also fixes a minor crash bug involving attempting to refresh the plots when no data are available.

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -62,28 +62,6 @@ static std::vector<PreparedData> ConvertToPrepared(
 }
 
 /**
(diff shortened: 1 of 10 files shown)


Oblarg:
image

calcmogul:
If I load a JSON, then click the analysis reset button, then try to choose "backward" in the direction combo box, it crashes. m_hasData isn't being set back to false when it should.

calcmogul (changes requested):
If I load a JSON, then click the analysis reset button, then try to choose "backward" in the direction combo box, it crashes. m_hasData isn't being set back to false when it should.

calcmogul:
We should look into refactoring the error handling at some point. We're approximating a state machine of some kind, but it's not formally defined, and we may not be catching all error conditions.

calcmogul (approved):

Piphi5 (approved):

PR #358, opened 2022-02-05
Refactor Analyzer into a state machine

### LINKED ISSUE #357: Refactor error handling
We should look into refactoring the error handling at some point. We're approximating a state machine of some kind, but it's not formally defined, and we may not be handling all error conditions properly. We'll need to enumerate all the error paths, the possible GUI operation modes, and what the mode transitions should be for each error.

For example, if loading a JSON fails, everything gets reset. If a JSON is loaded but it can't be processed (the data is malformed or some other random exception happens), throw the processed data away and just show the raw data. If no exceptions are thrown, the plots and analysis controls work.

Piphi5:
FIxes #357

### DIFF
--- a/sysid-application/src/main/native/include/sysid/view/Analyzer.h
+++ b/sysid-application/src/main/native/include/sysid/view/Analyzer.h
@@ -8,6 +8,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <string_view>
 #include <thread>
 #include <vector>
 
@@ -40,7 +41,22 @@ namespace sysid {
 class Analyzer : public glass::View {
  public:
   /**
-   * The different motor controlle rtiming presets that can be used.
+   * The different display and processing states for the GUI
+   */
+  enum class AnalyzerState {
+    kWaitingForJSON,
+    kNominalDisplay,
+    kDataPrep,
+    kFeedforwardGainCalc,
+    kFeedbackGainCalc,
+    kGraphPrep,
+    kMotionThresholdError,
+    kTestDurationError,
+    kGeneralDataError,
+    kFileError
+  };
+  /**
+   * The different motor controller timing presets that can be used.
    */
   static constexpr const char* kPresetNames[] = {"Default",
                                                  "WPILib (2020-)",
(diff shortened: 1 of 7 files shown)


sysid-application/src/main/native/include/sysid/view/Analyzer.h
```diff
@@ -40,7 +41,23 @@ namespace sysid {
 class Analyzer : public glass::View {
  public:
   /**
-   * The different motor controlle rtiming presets that can be used.
+   * The different display and processing states for the GUI
+   */
+  enum GUIState {
```
calcmogul, line 46: Use an `enum class` instead of an `enum` and wpilib's C++ enum value naming style. For example, `kWaitingForJSON` instead of `WAITING_FOR_JSON`.

sysid-application/src/main/native/include/sysid/view/Analyzer.h
```diff
@@ -128,21 +152,23 @@ class Analyzer : public glass::View {
...
-  /**
-   * Handles json errors when they pop up.
-   */
-  void HandleJSONError(const wpi::json::exception& e);
+  // State of the Display GUI
+  GUIState m_state;
+
+  // Stores the error display state the GUI must transition into after the error
+  // popup.
+  GUIState m_errorState;
```
calcmogul, line 162: Why is this variable needed in addition to m_state?

calcmogul (changes requested):

sysid-application/src/main/native/include/sysid/view/Analyzer.h
Piphi5: Yeah after further thought this wasn't needed

calcmogul:
The rest looks OK. I still need to check all the error paths. Should we commit some sample JSONs for each failure scenario so we can check those during review?

calcmogul (changes requested):
One thing I noticed when tweaking gearing in the feedback analysis section is that as soon as I type in a character, the GUI updates and it stops letting me enter text. That means I can't enter something like `8.16`. It assumes I'm done after `8`. This doesn't happen with main.

Piphi5:
> The rest looks OK. I still need to check all the error paths. Should we commit some sample JSONs for each failure scenario so we can check those during review?

Maybe we should. Only issue is that when we change the data collection most of them will become invalid.

Piphi5:
> One thing I noticed when tweaking gearing in the feedback analysis section is that as soon as I type in a character, the GUI updates and it stops letting me enter text. That means I can't enter something like `8.16`. It assumes I'm done after `8`. This doesn't happen with main.

This should be fixed now

calcmogul:
The sliders do it too. I can't drag the ones in the feedback section because the window scrolls back up to the top. I assume this is because the GUI widgets are disappearing during state transitions, then reappearing later.

Piphi5:
That means we'll probably have to split up the feedforward and feedback gain calculation. Probably shouldn't have grouped them to start with.

Piphi5:
So instead of if-else if statements, they're just all if statements so that we can just go through the processing statemachine and land back at the display state before the next display call.

calcmogul:
Sounds good. That seems to have fixed the GUI issues I was seeing.

Oblarg (approved):

Oblarg (changes requested):
Currently errors out if "reset" is selected.  Also occasionally crashes on startup after erroring out (following that it starts correctly).

calcmogul:
Here's the backtrace from GDB:
```
#0  0x000055555558a727 in sysid::Analyzer::DisplayFeedforwardParameters(float, float) ()
#1  0x000055555558aa5b in sysid::Analyzer::DisplayFeedforwardGains(float, float) ()
#2  0x000055555558d914 in sysid::Analyzer::Display() ()
#3  0x000055555560a083 in glass::Window::Display (this=0x555556d960d0)
    at /__w/allwpilib/allwpilib/glass/src/lib/native/cpp/Window.cpp:65
#4  0x00005555555ffe6c in glass::WindowManager::DisplayWindows (this=<optimized out>)
    at /__w/allwpilib/allwpilib/glass/src/lib/native/cpp/WindowManager.cpp:112
#5  0x0000555555621a0f in std::function<void ()>::operator()() const (this=<optimized out>)
    at /usr/include/c++/8/bits/std_function.h:260
#6  wpi::gui::CommonRenderFrame ()
    at /__w/allwpilib/allwpilib/wpigui/src/main/native/cpp/wpigui.cpp:357
#7  0x0000555555623d90 in wpi::gui::PlatformRenderFrame ()
    at /__w/allwpilib/allwpilib/wpigui/src/main/native/opengl3/wpigui_opengl3.cpp:81
#8  0x00005555556217dc in wpi::gui::Main ()
    at /__w/allwpilib/allwpilib/wpigui/src/main/native/cpp/wpigui.cpp:296
#9  0x00005555555a7345 in Application(std::basic_string_view<char, std::char_traits<char> >) ()
#10 0x00005555555895a8 in main ()
```

Piphi5:
This should fix the reset issue, what do you mean by crashing sometimes on startup @Oblarg

Oblarg:
@Piphi5 The screen would be blank except for the menu selections at the top, and would promptly freeze and require me to kill the process.

Piphi5:
Is there a way to reliably reproduce that?

Piphi5:
Also just noticed the analysis tests don't compile

Oblarg:
Working for me now.

Oblarg:
We might want to do some sanitizing on the JSON inputs; if you pass it a completely incorrect file (e.g. a generator settings file instead of a data json) the program crashes.

Oblarg (approved):

calcmogul (approved):

PR #362, opened 2022-02-07
Display right and left datasets

### LINKED ISSUE #361: Add back ability to generate left/right combined drivetrain feedforward values
#341 removed the combo box which selected the type of analysis to do on a drivetrain characterization dataset.  While I agree that it makes sense to remove the forward/backward confusion, I still think there is value in being able to obtain a separate left and right dataset.  I think the default value of the box could remain as "combined" but there is real value in being able to produce separate left and right (combined) values.

Piphi5:
Fixes #361

### DIFF
--- a/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
+++ b/sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
@@ -121,10 +121,10 @@ static Storage CombineDatasets(const std::vector<PreparedData>& slowForward,
  */
 static void PrepareGeneralData(
     const wpi::json& json, AnalysisManager::Settings& settings, double factor,
-    std::string_view unit, Storage& originalDataset, Storage& rawDataset,
-    Storage& filteredDataset, std::array<units::second_t, 4>& startTimes,
-    units::second_t& minStepTime, units::second_t& maxStepTime,
-    wpi::Logger& logger) {
+    std::string_view unit, std::array<Storage, 3>& originalDataset,
+    std::array<Storage, 3>& rawDataset, std::array<Storage, 3>& filteredDataset,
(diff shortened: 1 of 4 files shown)


sysid-application/src/main/native/cpp/analysis/AnalysisManager.cpp
```diff
@@ -179,12 +180,12 @@ static void PrepareGeneralData(
...
   WPI_INFO(logger, "{}", "Storing datasets.");
   // Store the raw datasets
-  rawDataset = CombineDatasets(
+  rawDataset[0] = CombineDatasets(
       preparedData["raw-slow-forward"], preparedData["raw-slow-backward"],
       preparedData["raw-fast-forward"], preparedData["raw-fast-backward"]);
 
   // Store the filtered datasets
-  filteredDataset = CombineDatasets(
+  filteredDataset[0] = CombineDatasets(
```
calcmogul, line 188: The enum should be used here instead of a magic number.

calcmogul (changes requested):

calcmogul:
Do you have a dataset I can test this PR against?

Piphi5:
Here are the datasets I used to double check. Granted, the drivetrain one isn't great but it at least showed that the values were different for the different datasets. And the other dataset I used to make sure it works fine for non-drive datasets.
Drivetrain:
[sysid_data20220122-085328.json.txt](https://github.com/wpilibsuite/sysid/files/8029639/sysid_data20220122-085328.json.txt)

Other:
[sysid_data20220114-202020.json.txt](https://github.com/wpilibsuite/sysid/files/8029641/sysid_data20220114-202020.json.txt)

calcmogul (approved):

Issue #367, opened 2022-02-12
“Reset Telemetry” resets logger mechanism type to Drivetrain


shueja-personal:
Clicking “Reset Telemetry” changes the logger mechanism type back to Drivetrain in addition to clearing run tests. This is unintuitive and results in tests being run with the wrong mechanism. Is it intentional?

PR #368, opened 2022-02-12
Don't reset selected project type on telemetry reset.

Oblarg:
Closes https://github.com/wpilibsuite/sysid/issues/367


sysid-application/src/main/native/cpp/view/Logger.cpp
```diff
@@ -51,7 +51,6 @@ void Logger::Display() {
   if (ImGui::Button("Reset Telemetry")) {
     m_settings = TelemetryManager::Settings{};
     m_manager = std::make_unique<TelemetryManager>(m_settings, m_logger);
-    m_selectedType = 0;
```
calcmogul, line 54: Should m_settings.mechanism be reinitialized like so?
```cpp
    m_settings.mechanism = analysis::FromName(kTypes[m_selectedType]);
```

calcmogul (approved):

PR #369, opened 2022-02-14
Add step test data collection procedure

### LINKED ISSUE #296: Implement stairstep data collection test
This would step the voltage by a predetermined amount, step it again when the velocity reaches steady-state, then repeat. A max distance could be implemented so the test knows when to start going backward. To determine steady-state, we could have a running variance of the acceleration based on code like this:
```cpp
double sysid::GetAccelNoiseFloor(const std::vector<PreparedData>& data,
                                 int window) {
  double sum = 0.0;
  size_t step = window / 2;
  auto averageFilter = frc::LinearFilter<double>::MovingAverage(window);
  for (size_t i = 0; i < data.size(); i++) {
    double mean = averageFilter.Calculate(data[i].acceleration);
    if (i >= step) {
      sum += std::pow(data[i - step].acceleration - mean, 2);
    }
  }
  return std::sqrt(sum / (data.size() - step));
}
```
where steady-state occurs when the acceleration drops below that noise floor.

This one test would replace all four of the existing tests since it sufficiently samples across the whole velocity-acceleration plane. Ks, Kv, and Ka are uniquely determined because Ka is determined by the transient, Ks manifests as a constant bias in the steady-state, and Kv manifests as a linearly increasing bias in the steady-state.

Piphi5:
Fixes #296 

- [ ] Implement noise floor calculation
- [ ] Figure out unit testing for sysid-library
- [ ] Figure out why certain tests don't compile


Piphi5:
Do we have a way to setup unit testing for the sysid-library? I think that was the major blocker for progress here.

calcmogul:
We should be able to add a test folder to sysid-library.

PR #372, opened 2022-02-17
Don't push NetworkTables entries in autonomous

calcmogul:
RobotPeriodic() is run after AutonomousPeriodic(), but it's still
nondeterminstic work we don't have to do for the actual test run.


Piphi5 (approved):
LGTM

PR #373, opened 2022-02-18
Read voltage measurements from motor controllers

### LINKED ISSUE #370: Voltage should be measured from the motor controller when possible
see discussion here: https://www.chiefdelphi.com/t/sysid-tool/398381/51?u=oblarg

Piphi5:
Fixes #370

### DIFF
--- a/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
+++ b/sysid-application/src/integrationtest-generation/native/cpp/GenerationTest.cpp
@@ -89,8 +89,8 @@ class GenerationTest : public ::testing::Test {
     std::this_thread::sleep_for(2s);
   }
 
-  void FindInLog(std::string_view str) {
-    auto searchString = fmt::format("Setup {}", str);
+  void FindInLog(std::string_view str, std::string_view prefix = "Setup") {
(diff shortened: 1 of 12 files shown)


Oblarg:
This should leverage the Rev/CTRE APIs where possible, as well.

Piphi5:
I can use GetBusVoltage right?

Oblarg:
I think?

Piphi5:
My only issue with the current version of this is that we don't have any good way of unit testing the if/else statements for each vendor motor. If we were to change the name of the motor controllers that could be a major bug that would go unnoticed.

I think the most ideal way of solving this is to use the hardware types class that the sysid application uses to avoid getting mixed up. However, I don't know enough gradle to get that to work.

Piphi5:
It would also be nice to have some sort of vendor verification that the methods here are the ones needed to get the proper voltage to motor reading.

Piphi5:
@CodingWithFusion can you verify if [this code](https://github.com/Piphi5/sysid/blob/fix-voltage-measurement/sysid-library/src/main/cpp/logging/SysIdLogger.cpp#L62) is correct for getting the voltage applied to the motor for the venom?

calcmogul (approved):

PR #374, opened 2022-02-19
Add theoretical mode

Oblarg:
Replaces the "waiting for data json" state with a theoretical analysis mode that allows you to plug in your own calculated feedforward gains for use with LQR.  Also refactors the state machine to get rid of states that do nothing but call code and transition to other states (these are now just subroutines).

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -42,12 +42,57 @@ Analyzer::Analyzer(glass::Storage& storage, wpi::Logger& logger)
   m_presets["REV Brushed Encoder Port"] = presets::kREVNonNEO;
   m_presets["REV Data Port"] = presets::kREVNonNEO;
   m_presets["Venom"] = presets::kVenom;
+
+  ResetData();
+  UpdateFeedbackGains();
 }
 
-void Analyzer::DisplayGain(const char* text, double* data) {
+void Analyzer::UpdateFeedforwardGains() {
+  WPI_INFO(m_logger, "{}", "Gain calc");
+  try {
+    const auto& [ff, trackWidth] = m_manager->CalculateFeedforward();
+    m_ff = std::get<0>(ff);
+    m_accelRSquared = std::get<1>(ff);
+    m_accelRMSE = std::get<2>(ff);
+    m_trackWidth = trackWidth;
+    PrepareGraphs();
+  } catch (const sysid::InvalidDataError& e) {
+    m_state = AnalyzerState::kGeneralDataError;
+    HandleError(e.what());
+  } catch (const sysid::NoQuasistaticDataError& e) {
+    m_state = AnalyzerState::kMotionThresholdError;
+    HandleError(e.what());
+  } catch (const sysid::NoDynamicDataError& e) {
+    m_state = AnalyzerState::kTestDurationError;
+    HandleError(e.what());
+  } catch (const AnalysisManager::FileReadingError& e) {
+    m_state = AnalyzerState::kFileError;
+    HandleError(e.what());
+  } catch (const wpi::json::exception& e) {
+    m_state = AnalyzerState::kFileError;
+    HandleError(e.what());
+  } catch (const std::exception& e) {
+    m_state = AnalyzerState::kFileError;
+    HandleError(e.what());
+  }
+}
+
+void Analyzer::UpdateFeedbackGains() {
+  const auto& fb = m_manager->CalculateFeedback(m_ff);
+  m_timescale = m_ff[2] / m_ff[1];
+  m_Kp = fb.Kp;
+  m_Kd = fb.Kd;
+}
+
+bool Analyzer::DisplayGain(const char* text, double* data,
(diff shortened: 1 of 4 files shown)


calcmogul:
The input boxes don't fit.
Screenshot_20220218_193448

Oblarg:
Should be fixed.  It'd be nice if I could reliably reset the formatting to default - we should have a button for that.

calcmogul:
The response timescale box is missing from the JSON analysis.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -323,6 +359,18 @@ void Analyzer::PrepareRawGraphs() {
...
 
+void Analyzer::PrepareGraphs() {
+  WPI_INFO(m_logger, "{}", "Graph state");
+  AbortDataPrep();
+  m_dataThread = std::thread([&] {
+    m_plot.SetData(m_manager->GetRawData(), m_manager->GetFilteredData(),
+                   m_manager->GetUnit(), m_ff, m_manager->GetStartTimes(),
+                   m_manager->GetAnalysisType(), m_abortDataPrep);
+  });
+  UpdateFeedbackGains();
```
Piphi5, line 370: Actually I see the issue, you call `UpdateFeedbackGains` inside `PrepareGraphs` in which `UpdateFeedbackGains` calls `PrepareGraphs` causing an infinite loop of graph preparation.

Piphi5 (changes requested):

sysid-application/src/main/native/cpp/view/Analyzer.cpp
Oblarg: I don't think this is true?  I don't see any call to `PrepareGraphs` from `UpdateFeedbackGains`, and it doesn't crash on my machine.

Piphi5: Oops yeah I misread

Piphi5:
It seems like data errors causes none of the graphs to be plotted.
image

Oblarg:
That is correct; I can change that if you want.

Piphi5:
Yeah ideally they can see the raw data so they can fix the motion threshold

Oblarg:
Is this good to merge?

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -86,21 +134,31 @@ void Analyzer::DisplayFileSelector() {
...
+  m_manager = std::make_unique<AnalysisManager>(m_settings, m_logger);
+  m_location = "";
+  m_ff = std::vector<double>{1, 1, 1};
+  UpdateFeedbackGains();
+}
+
 bool Analyzer::DisplayResetAndUnitOverride() {
+  auto type = m_manager->GetAnalysisType();
+  auto unit = m_manager->GetUnit();
+  auto factor = m_manager->GetFactor();
```
Piphi5, line 148: What's the reasoning behind getting rid of the member variables?

Piphi5 (commented):
One concern is that the weird value update bug happens if we load the file the second time:
I haven't checked if this happens on main but it's probably some logic we can fix on this PR.
https://user-images.githubusercontent.com/29788153/154989342-ba430192-55ff-410d-96d8-2f640336ca10.mp4

Edit: The video came out blurrier than I expected but essentially the first time around the calculated gain is at 2.0879 and then the second time around it starts at 2.0879 and then pressing on convert gains to encoder counts turns it into 2.7182

sysid-application/src/main/native/cpp/view/Analyzer.cpp
Oblarg: Duplicating state is never a good idea, because then you have to keep it synchronized.

Oblarg:
> One concern is that the weird value update bug happens if we load the file the second time: I haven't checked if this happens on main but it's probably some logic we can fix on this PR. https://user-images.githubusercontent.com/29788153/154989342-ba430192-55ff-410d-96d8-2f640336ca10.mp4

This happens in main immediately upon file load; I'm not sure why loading it twice would cause issues, though...

Oblarg:
I'm having a hard time replicating the bug at all, even on this branch - I think something might be funky here.

Oblarg:
Found the offending code, and it's fixed.  Should be good now.

Piphi5:
what was the error?

Oblarg:
Failing to early exit out of a data error state if the reset button is pressed.

Piphi5:
The error still shows up for me :/

Oblarg:
What dataset is causing it?  I cannot replicate it.

Piphi5:
The original one linked on the discord.

calcmogul:
> The original one linked on the discord.

It would be good to upload it here.

Piphi5:
[sysid_data20220215-175430.json.txt](https://github.com/wpilibsuite/sysid/files/8111137/sysid_data20220215-175430.json.txt)

Piphi5 (approved):

calcmogul (approved):

PR #380, opened 2022-02-21
Use text box instead of sliders for LQR params.

Oblarg:
Sliders have an arbitrary range which is not necessarily suitable for all units.


sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -751,28 +751,17 @@ void Analyzer::DisplayFeedbackGains() {
...
   }
 
-  ShowLQRParam(fmt::format("Max Velocity Error ({}/s)", abbreviation).c_str(),
-               &m_settings.lqr.qv, 0.05f, 40.0f);
-  ShowLQRParam("Max Control Effort (V)", &m_settings.lqr.r, 0.1f, 12.0f, false);
+  ImGui::SetCursorPosX(ImGui::GetFontSize() * 9);
+  DisplayGain(fmt::format("Max Velocity Error ({}/s)", abbreviation).c_str(),
+              &m_settings.lqr.qp, false);
+  ImGui::SetCursorPosX(ImGui::GetFontSize() * 9);
+  DisplayGain("Max Control Effort (V)", &m_settings.lqr.r, false);
```
Piphi5, line 766: Could we add a maximum voltage cap for 12 V? I think just a simple if statement setting m_settings.lqr.r to 12 if its greater than it should work.

Oblarg: I don't think there's really a reason to do this; it's fun to play with the tool to see what happens in the limiting cases, and you can always rescale both terms without changing the LQR results

Piphi5: Alright that's fair

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -751,28 +751,17 @@ void Analyzer::DisplayFeedbackGains() {
...
-                             float max, bool power = true) {
-    float val = *data;
-    ImGui::SetNextItemWidth(ImGui::GetFontSize() * 5);
-    ImGui::SetCursorPosX(ImGui::GetFontSize() * 9);
-
-    if (ImGui::SliderFloat(text, &val, min, max, "%.1f",
-                           ImGuiSliderFlags_None |
-                               (power ? ImGuiSliderFlags_Logarithmic : 0))) {
-      *data = static_cast<double>(val);
-      UpdateFeedbackGains();
```
Piphi5, line 764: Locally, changing the parameters doesn't change the gains. I think it's missing an `UpdateFeedbackGains` call somewhere.

Piphi5 (changes requested):

sysid-application/src/main/native/cpp/view/Analyzer.cpp
Oblarg: fixed

calcmogul (changes requested):
If you try to enter a number like 0.13, the GUI will recompute the gains after you enter 0 and crash. The input boxes should be changed to only take effect after enter is pressed.

Piphi5:
Is catching the exception in the scope of this PR though?

calcmogul:
Yes. The code on main never lets the tolerance be zero, so the exception was impossible to produce. This change makes it possible.

Oblarg:
Added a change that doesn't catch the exception but it makes it a *lot* harder to encounter.  I think this is good enough, and the exception handling at the calculation point should be a different PR.

calcmogul:
You need to add a guard for `m_settings.lqr.r` as well, because `1/0^2` breaks.

Oblarg:
Done.

calcmogul (approved):

PR #384, opened 2022-02-22
Add measurement delay estimation.

Oblarg:
Estimates the measurement delay for both position and velocity measurements from the step test data.

Position delays are measured from the time the driving voltage is applied, to the time the measured position exceeds `motion threshold * dt`.

Velocity delays are measured from the time the driving voltage is applied, to the time the computed acceleration reaches its maximum.

These values are now used by default in the feedback analyzer, but are overridden when a present is selected.

Selecting a new preset now resets the loop type to `velocity`, since the preset measurement delay only applies in that case.


sysid-application/src/main/native/include/sysid/analysis/AnalysisManager.h
```diff
@@ -267,15 +268,39 @@ class AnalysisManager {
...
+
+  /**
+   * Returns the estimated time delay of the measured velocity, including
+   * CAN delays.
+   *
+   * @return Velocity delay
+   */
+  units::second_t GetVelocityDelay() const {
+    return std::accumulate(m_velocityDelays.begin(), m_velocityDelays.end(),
+                           units::second_t{0}) /
```
calcmogul, line 301:
```suggestion
                           0_s) /
```

sysid-application/src/main/native/include/sysid/analysis/AnalysisManager.h
```diff
@@ -267,15 +268,39 @@ class AnalysisManager {
...
+
+  /**
+   * Returns the estimated time delay of the measured position, including
+   * CAN delays.
+   *
+   * @return Position delay
+   */
+  units::second_t GetPositionDelay() const {
+    return std::accumulate(m_positionDelays.begin(), m_positionDelays.end(),
+                           units::second_t{0}) /
```
calcmogul, line 289:
```suggestion
                           0_s) /
```

calcmogul (changes requested):
Since the delays are almost always below 1 second, it would be easier to read them if they were in units of milliseconds.

Oblarg:
we'll do the milliseconds change in a separate PR, since it hits more things

calcmogul (approved):

PR #386, opened 2022-02-25
Fix units not displaying after data collection

prateekma:


calcmogul:
The cause seems to be that std::transform() stops iterating over the destination range if it's too short. https://godbolt.org/z/qr3fdsKs8

calcmogul (approved):

PR #395, opened 2022-03-05
Only graph when data is present

### LINKED ISSUE #393: Diagnostic plot point size causes crash with zero data
Steps to reproduce
1. Install WPILIb 2022.4.1 on a new system
2. Open SysID from Start (No data should be loaded)
4. Go to the Diagnostic Plots window and adjust the point size slider.

This particular system is running Win11 21H2, but this also happens on both Win10 20H1 and 21H2. No error message is returned, the GUI just closes.

Piphi5:
Fixes #393 

Previously changing the graphing settings before any data was present would lead to it crashing.

### DIFF
--- a/sysid-application/src/main/native/include/sysid/analysis/AnalysisManager.h
+++ b/sysid-application/src/main/native/include/sysid/analysis/AnalysisManager.h
@@ -307,7 +307,15 @@ class AnalysisManager {
    *
    * @return The start times for each test
    */
-  const std::array<units::second_t, 4> GetStartTimes() { return m_startTimes; }
+  const std::array<units::second_t, 4>& GetStartTimes() const {
+    return m_startTimes;
+  }
+
+  bool HasData() const {
+    return !m_originalDataset[static_cast<int>(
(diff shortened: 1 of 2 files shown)


sysid-application/src/main/native/include/sysid/analysis/AnalysisManager.h
```diff
@@ -309,6 +309,12 @@ class AnalysisManager {
    */
   const std::array<units::second_t, 4> GetStartTimes() { return m_startTimes; }
 
+  const bool HasData() {
+    return !m_originalDataset[static_cast<int>(
+                                  Settings::DrivetrainDataset::kCombined)]
+                .empty();
+  }
```
calcmogul, line 316:
```suggestion
  const std::array<units::second_t, 4>& GetStartTimes() const { return m_startTimes; }

  bool HasData() const {
    return !m_originalDataset[static_cast<int>(
                                  Settings::DrivetrainDataset::kCombined)]
                .empty();
  }
```

calcmogul (changes requested):

calcmogul (approved):

PR #397, opened 2022-03-05
Change some time constants to milliseconds

Oblarg:
Also adds a guard for divide-by-zero when when using kv/ka to determine a time constant.

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -82,10 +82,12 @@ void Analyzer::UpdateFeedforwardGains() {
 }
 
 void Analyzer::UpdateFeedbackGains() {
-  const auto& fb = m_manager->CalculateFeedback(m_ff);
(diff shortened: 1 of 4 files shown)


sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -547,7 +549,8 @@ void Analyzer::CollectFeedforwardGains(float beginX, float beginY) {
   SetPosition(beginX, beginY, 0, 2);
   // Show Timescale
   ImGui::SetNextItemWidth(ImGui::GetFontSize() * 4);
-  DisplayGain("Response Timescale (s)", &m_timescale);
+  DisplayGain("Response Timescale (s)",
```
calcmogul, line 552:
```suggestion
  DisplayGain("Response Timescale (ms)",
```

calcmogul (changes requested):

calcmogul (approved):

calcmogul (approved):

PR #405, opened 2022-03-18
Allow user to set 0 measurement delay

Oblarg:
The check was previously >0 rather than >=0.

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Analyzer.cpp
+++ b/sysid-application/src/main/native/cpp/view/Analyzer.cpp
@@ -682,14 +682,17 @@ void Analyzer::DisplayFeedbackGains() {
     }
 
     ImGui::SetNextItemWidth(ImGui::GetFontSize() * 4);
-    if (ImGui::InputDouble(text, data, 0.0, 0.0, "%.5G") && *data > 0) {
-      UpdateFeedbackGains();
-    }
+    return ImGui::InputDouble(text, data, 0.0, 0.0, "%.5G");
   };
 
   // Show controller period.
-  ShowPresetValue("Controller Period (ms)",
-                  reinterpret_cast<double*>(&m_settings.preset.period));
+  if (ShowPresetValue("Controller Period (ms)",
+                      reinterpret_cast<double*>(&m_settings.preset.period))) {
+    if (m_settings.preset.period > 0_s &&
+        m_settings.preset.measurementDelay >= 0_s) {
+      UpdateFeedbackGains();
+    }
+  }
 
   // Show whether the controller gains are time-normalized.
   if (ImGui::Checkbox("Time-Normalized?", &m_settings.preset.normalized)) {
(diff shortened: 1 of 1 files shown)


calcmogul (changes requested):
A lot of these changes formatting changes don't seem related to the PR, and wpiformat is complaining about some of them.

Oblarg:
Do I have the wrong version of wpiformat, somehow?

calcmogul:
Could be a different version of clang-format. wpiformat doesn't do the actual formatting itself.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -682,26 +682,34 @@ void Analyzer::DisplayFeedbackGains() {
...
+          reinterpret_cast<double*>(&m_settings.preset.measurementDelay))) {
+    if (m_settings.preset.measurementDelay >= 0_s) {
+      UpdateFeedbackGains();
+    }
+  }
+
   sysid::CreateTooltip(
-      "The average measurement delay of the process variable in milliseconds. "
+      "The average measurement delay of the process variable in "
+      "milliseconds. "
```
calcmogul, line 712: This change seems extraneous.

Oblarg:
Seems fixed now.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
```diff
@@ -754,8 +762,10 @@ void Analyzer::DisplayFeedbackGains() {
       UpdateFeedbackGains();
     }
     sysid::CreateTooltip(
-        "The counts per rotation of your encoder. This is the number of counts "
-        "reported in user code when the encoder is rotated exactly once. Some "
+        "The counts per rotation of your encoder. This is the number of "
+        "counts "
+        "reported in user code when the encoder is rotated exactly once. "
+        "Some "
```
calcmogul, line 768: This change seems extraneous.

sysid-application/src/main/native/cpp/view/Analyzer.cpp
Oblarg: yeah I think those reflowed from a previous WPIformat, let me undo them and see if it works

calcmogul (changes requested):
The controller period starts as nonzero and the measurement delay starts as zero. If you set the controller period to zero then change the measurement delay to any value >= 0, the tool tries to update the feedback gains and the tool crashes with the following exception.
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  The system passed to the LQR is uncontrollable!

A =
  1.000000
B =
  0.000000

Aborted (core dumped)
```
UpdateFeedbackGains() should only be called when one of the values changes and both values are in a valid range, not just when one of them is. This bug doesn't occur on main because both gains are required to be greater than zero.

calcmogul (approved):

PR #426, opened 2022-04-25
Fix Override Units not overriding units per rotation

### LINKED ISSUE #404: Override Units doesn't override units per rotation

Piphi5:
Fixes #404


calcmogul (approved):

calcmogul:
Previously, Override Units would replace whatever the user entered for the conversion factor with 1.0. This fixed that.

PR #432, opened 2022-05-06
Assert wpilib version

### LINKED ISSUE #431: Assert Wpilib version = 2022.4.1
Recent updates to wpilib have created breaking changes in vendor libraries - namely, Phoenix and PlayingWithFusion2022. In order to fix linker issues, we should set the wpilib version to the latest version that works with vendor libraries.

jlmcmchl:
Fixes #431

### DIFF
--- a/sysid-projects/analysis-test/src/main/include/Elevator.h
+++ b/sysid-projects/analysis-test/src/main/include/Elevator.h
@@ -77,6 +77,5 @@ class Elevator : public SysIdGeneralMechanism {
                                             1000,
                                             2_in,
                                             -Constants::Elevator::kHeight,
-                                            Constants::Elevator::kHeight,
-                                            true};
+                                            Constants::Elevator::kHeight};
 };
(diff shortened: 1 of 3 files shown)


sysid-projects/analysis-test/src/main/include/Elevator.h
```diff
@@ -77,6 +77,5 @@ class Elevator : public SysIdGeneralMechanism {
                                             1000,
                                             2_in,
                                             -Constants::Elevator::kHeight,
-                                            Constants::Elevator::kHeight,
-                                            true};
+                                            Constants::Elevator::kHeight};
```
prateekma, line 80: This shouldn't have been changed (not in this PR at least).

prateekma (changes requested):

sysid-projects/analysis-test/src/main/include/Elevator.h
jlmcmchl: It fixed the build on my end, since elevator sim has added the option to simulate gravity since release 2022.4.1. I can take it out, though.

prateekma: Ah, didn't realize that. Then it should be fine.

prateekma (approved):

Issue #406, opened 2022-03-19
Add CANivore support


calcmogul:
To support CANivore, we'd need to add an optional CANivore bus string field to the GUI for the CTRE devices that support it (not all of them do). An empty string would use the original constructor they use now.

It requires conditionally including a textbox, adding an entry for it in the JSON, and modifying the robot programs to use that field conditionally.

ThadHouse:
Isnt there a sentinel value to use the roborio bus? So you wouldn't have to use an alternate constructor. And could just make the default in the UI be whatever that sentinel is.

Piphi5:
I think passing in `rio` defaults it to the roborio according to [here](https://docs.ctre-phoenix.com/en/stable/ch08a_BringUpCANivore.html#canivore-api).

Piphi5:
Since this will most likely be a 2023 change and will probably change the current hardware config json file format, would we want the currently saved hardware config files from this year to be compatible after the CANivore changes?

calcmogul:
SysId should be able to open old generator JSONs and autoconvert them to the new format, but the robot program doesn't need to be able to read the old format.

calcmogul:
Closed by #455.

PR #455, opened 2022-12-21
Add CTRE CANivore and Phoenix Pro support

TytanRock:
Adds support for CANivore and Phoenix Pro devices.

Currently known bugs:
 - [x] Feedforward analysis gains are off by a factor of 12


sysid-library/src/main/include/sysid/generation/SysIdSetup.h
```diff
@@ -134,9 +141,11 @@ void SetupGyro(
     std::unique_ptr<frc::Gyro>& gyro,
     std::unique_ptr<frc::ADIS16448_IMU>& ADIS16448Gyro,
     std::unique_ptr<frc::ADIS16470_IMU>& ADIS16470Gyro,
-    // std::unique_ptr<BasePigeon>& pigeon,
-    // std::unique_ptr<WPI_TalonSRX>& tempTalon,
-    std::function<double()>& gyroPosition, std::function<double()>& gyroRate);
+    std::unique_ptr<BasePigeon>& pigeon,
+    std::unique_ptr<ctre::phoenixpro::hardware::Pigeon2>& pigeonpro,
+    std::unique_ptr<WPI_TalonSRX>& tempTalon,
+    const std::string& gyroCANivoreName, std::function<double()>& gyroPosition,
```
calcmogul, line 147: An `@param` doc is missing for gyroCANivoreName.

sysid-application/src/main/native/cpp/generation/ConfigManager.cpp
```diff
@@ -38,6 +41,13 @@ wpi::json ConfigManager::Generate(size_t occupied) {
   }
   json["motor controllers"] = motorControllers;
 
+  // Add CANivore busses
+  canivoreNames.clear();
+  for (size_t i = 0; i < occupied; i++) {
+    canivoreNames.push_back(std::string(m_config.canivoreNames[i].data()));
+  }
+  json["canivore names"] = canivoreNames;
+
```
Piphi5, line 50: Don't forget that if you're going to add new keys to the JSON that you need to update the `ReadJSON` method to make sure that it's in the json keys array and that the config is updated when its read.

sysid-application/src/main/native/cpp/view/Generator.cpp
```diff
@@ -346,6 +356,15 @@ void Generator::Display() {
     ImGui::SameLine();
     ImGui::Checkbox(drive ? "L Inverted" : "Inverted", &pi[i]);
 
+    if (m_settings.motorControllers[i] == sysid::motorcontroller::kTalonFX ||
+        m_settings.motorControllers[i] == sysid::motorcontroller::kTalonFXPro) {
+      // Add CANivore name if we are using a CTRE motor controller
+      ImGui::SameLine();
+      ImGui::SetNextItemWidth(80);
+      ImGui::InputText("MC CANivore Name", m_settings.canivoreNames[i].data(),
+                       m_settings.canivoreNames[i].size());
```
Piphi5, line 365: Not too familiar with the CANivore, but is it specific to only the Falcons?

Piphi5 (changes requested):

sysid-application/src/main/native/cpp/view/Generator.cpp
TytanRock: CANivore is specific to Falcons, CANcoder, and Pigeon2, which is why I have a field for all three of those devices.
https://github.com/wpilibsuite/sysid/pull/455/files#diff-4db26c65af78f49cb0a7385146e99aa2e25dcc97f6c52196d72c8bd8d5dff015R518
https://github.com/wpilibsuite/sysid/pull/455/files#diff-4db26c65af78f49cb0a7385146e99aa2e25dcc97f6c52196d72c8bd8d5dff015R148

It appears if I have multiple fields with the same name in ImGui, they tie to the same value. So I had to make their names unique since in concept you can have multiple CANivores on your robot, or at least separate devices between the CANivore bus and the RIO bus.
That said, I made the assumption that drivetrain Falcons are on the same bus, whether that is RIO or CANivore, which I think is a safe assumption since it seems odd to have half your drivetrain on one bus and the other half on another bus.
I'm also not tied to the current names in ImGui, so if you have suggestions on what to name the fields I'm all ears!

sysid-application/src/main/native/cpp/generation/ConfigManager.cpp
TytanRock: Is the `is drivetrain` key intentionally omitted from the ReadJSON method? I noticed that's missing in my review of the method.

calcmogul:
The latest formatting failure may be from not using clang-format 14 (what CI uses).

TytanRock:
I noticed during the CANcoder Pro addition that `m_encoderIdx` in `Generator.cpp` isn't enumerated, since it relies on getting the index of an array of strings. This made the CANcoder Pro addition risky, since I just looked at wherever the index was used to call `CANCoderSetup` and added the next index. Maybe in the future this can be refactored to make additions easier?

For now I'd like extra scrutiny on the `Generator.cpp` changes to make sure I didn't miss anything. Based on my local run it looks all right.

sciencewhiz:
Fixes #406

sysid-library/src/main/cpp/logging/SysIdLogger.cpp
```diff
@@ -51,6 +52,13 @@ double SysIdLogger::MeasureVoltage(
       if constexpr (frc::RobotBase::IsSimulation()) {
         fmt::print("Recording SPARK MAX voltage\n");
       }
+    } else if (controllerNames[i] == "TalonFX (Pro)") {
+      auto* ctreController =
+          dynamic_cast<ctre::phoenixpro::hardware::TalonFX*>(controller);
+      sum += ctreController->GetDutyCycle().GetValue().value() * 12.0;
```
calcmogul, line 58: It would be more accurate to record the applied voltage instead of multiplying the duty cycle by 12.

calcmogul (changes requested):

calcmogul (approved):

calcmogul:
Is this good to go or is it waiting on a final test run?

TytanRock:
Final test run shows a kV of 0.11407 which is roughly what I'd expect, so I'd say this is good to merge.

PR #478, opened 2023-02-05
Make Spark Max built-in encoder sampling and measurement period configurable

calcmogul:
The number of samples must be between 8 and 64 inclusive.

### DIFF
--- a/sysid-application/src/main/native/cpp/view/Generator.cpp
+++ b/sysid-application/src/main/native/cpp/view/Generator.cpp
@@ -4,6 +4,7 @@
 
 #include "sysid/view/Generator.h"
 
+#include <algorithm>
 #include <memory>
 #include <mutex>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


sysid-application/src/main/native/cpp/view/Generator.cpp
```diff
@@ -459,17 +460,23 @@ void Generator::Display() {
...
         m_settings.encoderType == sysid::encoder::kCANcoderPro)) {
     // Samples Per Average Setting
     ImGui::SetNextItemWidth(ImGui::GetFontSize() * 2);
-    ImGui::InputInt("Samples Per Average", &m_settings.numSamples, 0, 0);
+    if (ImGui::InputInt("Samples Per Average", &m_settings.numSamples, 0, 0)) {
+      // Spark Max built-in encoder number of samples must be within 8 to 64
+      // inclusive
+      if (mainMotorController == sysid::motorcontroller::kSPARKMAXBrushless &&
+          m_settings.encoderType == sysid::encoder::kSMaxEncoderPort) {
+        m_settings.numSamples = std::clamp(m_settings.numSamples, 8, 64);
```
Starlight220, line 477: Do we want to clamp or reject out-of-bounds here?

sysid-application/src/main/native/cpp/view/Generator.cpp
```diff
@@ -459,17 +460,23 @@ void Generator::Display() {
...
+      }
+    }
     CreateTooltip(
         "This helps reduce encoder noise by averaging collected samples "
         "together. A value from 5-10 is reccomended for encoders with high "
```
Starlight220, line 482: Does this recommendation tooltip need modification?

sysid-application/src/main/native/cpp/view/Generator.cpp
calcmogul: As a rule, we clamp invalid inputs. To restore the original setting, we'd have to store what the previous value was.

rzblue:
Oops, that restriction applies to the measurement period. The depth has a different requirement

calcmogul:
I pushed update UI configuration for the number of samples and measurement period, but I don't know if it's correct. The NEO built-in encoder was never exposed, as far as I can tell, since "built-in" wasn't in the list of encoders. Is that what "encoder port" is supposed to do?

PR #518, opened 2023-09-20
Remove project generator

### LINKED ISSUE #516: Remove project generator
It's too difficult to maintain all the combinations of hardware in the project generator, and the robot program we deploy has NT connection issues a lot of the time. We should provide logging code in WPILib for users to use in their own robot projects instead.

The first step is removing project generator support. See also the issue for adding a user-side logging class: #517.

Since vendordeps won't be included anymore, we should be able to migrate SysId into the allwpilib monorepo to streamline releases.

calcmogul:
Fixes #516.


calcmogul:
The integration tests were all removed since there's no simulated process for the logger to talk to anymore. The old tests were flaky anyway and need to be rewritten; most of them were already disabled in CI.

